PR #184, opened 2025-11-07
fix(sentinel): retry on conn reset, one more time

vihu:
# Summary

Fix https://github.com/malbeclabs/infra/issues/346

Fixes connection reset retry logic that was added in #177 but didn't work properly.

- The previous fix attempted to detect "connection reset" errors by checking if `reqwest_err.to_string().contains("connection reset")`, but this never matched because `Display` for `Kind::Request` only outputs `"error sending request"` without the nested source details
- This caused connection reset errors to bypass retry logic and immediately log at ERROR level, triggering Grafana alerts
- This implementation walks the error source chain (following the same pattern as `reqwest::Error::is_timeout()`, which you can refer [here](https://github.com/seanmonstar/reqwest/blob/a97e1956dd14a79b0207082e327098331519bf2b/src/error.rs#L116)) to find the underlying `io::Error` and check for `io::ErrorKind::ConnectionReset`

### DIFF
--- a/crates/sentinel/CHANGELOG.md
+++ b/crates/sentinel/CHANGELOG.md
@@ -7,7 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- retry on ECONNRESET [#177](https://github.com/doublezerofoundation/doublezero-offchain/pull/177))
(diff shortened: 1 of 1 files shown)


karl-dz (approved):

karl-dz (changes requested):
Can you update the changelog?

PR #198, opened 2025-11-17
fix(contributor-rewards): make scheduler retry infinitely

vihu:
# Summary

contributor-rewards scheduler was entering a permanent failure state after 10 consecutive RPC failures, requiring manual intervention to recover. This PR
removes the hard halt and implements self-healing retry logic with monitoring alerts.

Fixes https://github.com/malbeclabs/infra/issues/368

## Changes

- Remove hard halt: Scheduler now retries indefinitely instead of halting after max failures
- Add RPC retry with backoff: Wrap get_epoch_info() with backoff
- Add periodic alerting: Emit ERROR log every 10 consecutive failures for monitoring
- Remove max_consecutive_failures config: No longer needed since hard halt is removed


ben-dz (approved):
I think this looks good to me.

PR #199, opened 2025-11-18
feat(scheduler): automate calculate distribution

bgm-malbeclabs:
## Summary of Changes
This PR adds in a worker to automatically run calculate_distribution. It follows the existing pattern of having a light interface into the underlying validator-debt crate and runs every two hours. It works by getting the current dz_epoch, then subtracting one epoch to get the most recently completed epoch.  If the epoch is already finalized, the worker just shuts down. Otherwise it continues its run as usual. In a subsequent PR, we'll add a counter to this that must be reached before the debt is finalized. 

Note that due to the differences in floating point math (how "society" rounds - round 5 up) and scalar math (safer with integers - floor) there can be a 1 lamport difference between Nihar's CSV and the calculations provided by the `mul_scalar/1` method that Karl created. We've verified for every epoch that there's no greater than 1 lamport difference between the csv and the calculations in the validator-debt crate.

Closes https://github.com/malbeclabs/doublezero/issues/2027

## Testing Verification

This has been tested and results posted in [#tmp-validator-debt](https://malbeclabs.slack.com/archives/C09LES1Q127/p1763490621115539)


karl-dz (changes requested):
Fixing PR title

Issue #155, opened 2025-10-16
Add validation checks to request-validator-access


juan-malbeclabs:
Currently, the `request-validator-access` command does not perform the same pre-checks as `prepare-validator-access`. As a result, if any required argument or configuration is incorrect, the transaction is still sent on-chain and ends up being rejected.

We should:

1. Add the same validation logic from prepare-validator-access into request-validator-access, so that incorrect inputs stop execution before sending the transaction.
2. Introduce a new -f (force) flag that allows the transaction to be sent anyway. In this mode, the errors will be shown only as warnings, and the transaction will proceed regardless.

Benefits:

- Prevents unnecessary rejected transactions on-chain.
- Gives operators the flexibility to override checks when needed.

Acceptance Criteria:

- `request-validator-access` runs all the same validation checks as prepare-validator-access.
- Without -f, execution stops on validation errors.
- With -f, validation errors are downgraded to warnings, and the transaction is still submitted.

juan-malbeclabs:
Yes, the idea is that before making the request, it should validate that it’s feasible to approve it.

karl-dz:
Addressed in https://github.com/doublezerofoundation/doublezero-offchain/pull/211

PR #211, opened 2025-11-24
Add shared validator access validation and force flag

RedaRahmani:
## Summary of Changes

This PR implements the behavior described in #155 for the passport CLI:

- Extracts the validator access validation (gossip + leader scheduler checks for
  primary and backup validators) into a shared helper module:
  `crates/solana-cli/src/command/passport/access_validation.rs`.
- Refactors `prepare-validator-access` to use this shared helper, preserving the
  existing output and `--force` semantics.
- Extends `request-validator-access` to:
  - print the same cluster / validator info as `prepare-validator-access`,
  - run the same validation checks before PDA/signature/tx submission, and
  - support a new `-f/--force` flag so operators can override validation while
    still seeing the errors.
- Wires the new module into `crates/solana-cli/src/command/passport/mod.rs`.

With this change, `request-validator-access` now mirrors the validation behavior
of `prepare-validator-access`:
- Without `--force`: validation errors stop execution before the transaction is sent.
- With `--force`: validation errors are printed as warnings and the transaction
  is still submitted.

## Testing Verification

I ran the CLI crate tests locally:

```bash
cargo test -p doublezero-solana-cli

### DIFF
--- a/crates/solana-cli/src/command/passport/access_validation.rs
+++ b/crates/solana-cli/src/command/passport/access_validation.rs
@@ -1,5 +1,7 @@
 use anyhow::{Result, bail};
-use doublezero_ledger_sentinel::client::solana::SolRpcClientType;
+use doublezero_ledger_sentinel::{
+    client::solana::SolRpcClientType, constants::ENV_PREVIOUS_LEADER_EPOCHS,
+};
 use doublezero_solana_client_tools::rpc::SolanaConnection;
 use solana_client::rpc_response::RpcContactInfo;
 use solana_sdk::pubkey::Pubkey;
@@ -11,7 +13,7 @@ pub async fn validate_validator_access<C>(
     sol_client: &C,
     primary_validator_id: &Pubkey,
     backup_validator_ids: &[Pubkey],
-    leader_schedule_epochs: u8,
+    leader_schedule_epochs: Option<u8>,
 ) -> Result<Vec<String>>
 where
     C: SolRpcClientType + Sync,
@@ -36,12 +38,13 @@ pub async fn validate_validator_access_with_nodes<C>(
     sol_client: &C,
     primary_validator_id: &Pubkey,
     backup_validator_ids: &[Pubkey],
-    leader_schedule_epochs: u8,
+    leader_schedule_epochs: Option<u8>,
 ) -> Result<Vec<String>>
 where
     C: SolRpcClientType + Sync,
 {
     let mut errors = Vec::<String>::new();
+    let leader_schedule_epochs = leader_schedule_epochs.unwrap_or(ENV_PREVIOUS_LEADER_EPOCHS);
 
     println!("Primary validator 🖥️  💎:\n  ID: {primary_validator_id} ");
     if let Some(node) = find_node_by_node_id(nodes, primary_validator_id) {
@@ -129,7 +132,9 @@ pub fn should_continue_after_validation(errors: &[String], force: bool) -> bool 
 mod tests {
     use std::net::{Ipv4Addr, SocketAddr};
 
-    use doublezero_ledger_sentinel::client::solana::MockSolRpcClientType;
+    use doublezero_ledger_sentinel::{
+        client::solana::MockSolRpcClientType, constants::ENV_PREVIOUS_LEADER_EPOCHS,
+    };
     use solana_client::rpc_response::RpcContactInfo;
     use solana_sdk::pubkey::Pubkey;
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: crates/solana-cli/src/command/passport/request_access.rs [hunk 1]
```diff
@@ -127,3 +187,32 @@ impl RequestValidatorAccessCommand {
...
+    use crate::command::passport::access_validation::should_continue_after_validation;
+
+    #[test]
+    fn request_access_aborts_when_validation_fails_without_force() {
+        let errors = vec!["validation failed".to_string()];
+        let mut request_invoked = false;
+
+        if should_continue_after_validation(&errors, false) {
+            request_invoked = true;
+        }
```
karl-dz:
```suggestion
        let request_invoked = should_continue_after_validation(&errors, false);
```

crates/solana-cli/src/command/passport/access_validation.rs
```diff
@@ -0,0 +1,252 @@
...
+pub fn print_validation_errors(errors: &[String]) {
+    if errors.is_empty() {
+        return;
+    }
+
+    println!("\nErrors found:");
+    for error in errors {
+        println!(" - {}", error);
+    }
+}
```
karl-dz: You can inline the printing in `should_continue_after_validation`. I don't see anything else using `print_validation_errors`

karl-dz (changes requested):
Thank you for your contribution. I left a few comments

RedaRahmani:
Thanks a lot for the detailed review, @karl-dz ! 

I’ve updated the PR to:
- move `request_access` back to an instance method on `RequestValidatorAccessCommand`,
- hide the `--force` flag via `#[arg(hide = true)]`,
- inline the validation error printing into `should_continue_after_validation`,
- and remove the duplicate tests from `request_access.rs` (tests live in `access_validation.rs` now).
If there’s anything else you’d like me to adjust or clean up, I’m happy to follow up.

karl-dz (changes requested):
Very nice, thank you for making those changes.

If you can also add an entry to the CHANGELOG.md file in _crates/solana-cli_ like how the other entries are (you can put it under "Unreleased"), I can go ahead and approve this guy after the checks pass

RedaRahmani:
Hi @karl-dz , I rebased feat/passport-validator-access-validation onto upstream/main, resolved the CHANGELOG.md conflict, and force-pushed. The commits 94fa4c4 and 420e06f are no longer in the branch history. Please let me know if anything else looks off or if you’d like me to tweak the approach.

karl-dz:
> Hi @karl-dz
> 
> I’ve fixed the formatting issues (`just fmt-check` is now clean).
> 
> I also tried running `sh/test_doublezero_solana_clean.sh` locally. The script currently fails at `doublezero-solana passport fetch -ul --config` with:
> 
> `Error: Failed to fetch account F5hVnYTU4CeSxhyGQBKHQRZFtoCn5NBUNWMXP3T7giAR`.
> 
> This seems related to the local fork / config account rather than the new validator access validation logic, but I’m happy to adjust the script if you’d like me to change how the test is set up.

I'll have to put up a PR to remove that shell script. Try to follow the workflow found in local-validator.yml: https://github.com/RedaRahmani/doublezero-offchain/blob/feat/passport-validator-access-validation/.github/workflows/local-validator.yml#L37-L42. Let me know if you have an issue using `doublezero-solana-fork`.

karl-dz:
> Hi @karl-dz , Thanks a lot for the detailed review!
> 
> I’ve re-run sh/test_doublezero_solana_clean.sh on my side following the `local-validator.yml` workflow (using `doublezero-solana-fork` with `-um --reset --god-mode)`.
> 
> With that setup:
> 
> `doublezero-solana passport fetch -ul --config` works and prints the program config correctly. I couldn’t reproduce the Failed to fetch account F5hV… error once I followed the same steps as in local-validator.yml.
> 
> The part that now fails without any extra flags is
> 
> `doublezero-solana passport request-validator-access ...`
> 
> because the new validation correctly complains that the local validator is not an active staked validator, so the `access-request account` is never created. The final
> 
> `doublezero-solana passport fetch -ul --access-request <service_key>`
> 
> then fails with Failed to `fetch account …`, which is expected given the stricter checks.
> 
> If I add the new `--force` flag to the `request-validator-access` call in the script, the whole flow passes: the access request is created and the last fetch `--access-request` succeeds. So behaviour-wise the CLI changes seem correct; it’s really the shell script that needs to be adapted for the “no stake on local fork” case.
> 
> I’m happy to:
> 
> update `sh/test_doublezero_solana_clean.sh` in this PR to include `--force` for the local-validator scenario, or
> 
> drop the shell script and rely purely on the local-validator.yml workflow as you suggested.
> 
> Let me know which option you’d prefer and I’ll update the PR accordingly. Could you also please confirm that I’m validating these new checks in the right way, or let me know if there’s any additional scenario you’d like me to cover?

Let me try something locally. Will keep you updated.

karl-dz (approved):
Thank you for addressing this issue!

PR #215, opened 2025-11-28
Handle paginated Jito rewards and missing entries

RedaRahmani:
## Summary of Changes

This PR fixes an under-accounting issue in the Jito MEV rewards used by `validator-debt`.

Previously, `get_jito_rewards` fetched Jito rewards with a single request:

- It called `validator_rewards?epoch={epoch}&limit={JITO_REWARDS_LIMIT}`
- It assumed all rewards fit into one response, even when `total_count` exceeded the `limit`
- As a result, only the first page of results was used, and any additional rewards on later pages were silently ignored

This meant that for epochs where Jito returned more entries than the configured limit, validator MEV revenue could be underestimated and downstream debt calculations would be based on incomplete data.

This PR:

- Implements proper pagination for the Jito `validator_rewards` API
- Adds a safety cap (`MAX_JITO_PAGES`) to prevent runaway pagination if the API misbehaves
- Aggregates MEV revenue per validator across all fetched pages
- Improves logging around pagination and retries
- Extends tests to cover pagination and the “missing Jito entry” case

### Jito rewards fetching

In `crates/validator-debt/src/jito.rs`:

- Switched from a single-request flow to a paginated loop using:
  - `page` parameter
  - `limit={JITO_REWARDS_LIMIT}`
- Track:
  - `total_count` from the API
  - `collected` number of rewards across all pages
- Continue fetching pages until one of the following conditions is met:
  - We have collected at least `total_count` entries
  - We hit an empty page (defensive stop)
  - We reach `MAX_JITO_PAGES` (safety cap), in which case we return partial results with a log message

Additional improvements:

- Use a reusable exponential backoff retry strategy for each page fetch
- Upgrade error handling to use `anyhow::Context` so failures clearly indicate:
  - Which epoch failed
  - Which page failed
  - That retries were exhausted
- Aggregate rewards into a `HashMap<String, u64>` so that if a validator appears on multiple pages, their `mev_revenue` is summed
- Convert the aggregated map back into `HashMap<&str, u64>` keyed by the input `validator_ids`, defaulting to `0` when a validator has no Jito entry

### Reward aggregation

In `crates/validator-debt/src/rewards.rs`:

- `get_total_rewards` still combines:
  - Inflation rewards
  - Jito rewards
  - Block base + priority fees
- Behavior is now explicitly validated for the case where:
  - A validator has block and inflation rewards
  - But **no** Jito entry is returned
- In that case, the Jito component is treated as `0`, and the total is computed as:
  - `total = block_base + block_priority + inflation + jito (0)`

This makes the expected behavior around “missing Jito entries” explicit and tested.

### Unit tests

Added and updated tests in `crates/validator-debt/src/jito.rs`:

- `get_jito_rewards_handles_single_page_when_total_under_limit`
  - Verifies behavior when `total_count` fits within a single page
- `get_jito_rewards_fetches_all_pages_when_total_exceeds_page_size`
  - Uses a `Sequence` to ensure multiple pages are fetched in order
  - Asserts that rewards from all pages are correctly merged
- `get_jito_rewards_stops_on_empty_page`
  - Verifies that an empty page stops pagination and that missing validators get `0` as Jito rewards

Extended tests in `crates/validator-debt/src/rewards.rs`:

- Updated existing tests to account for the new paginated Jito API shape (`total_count` and `page=1` in URLs)
- Added:
  - `test_get_total_rewards_treats_missing_jito_as_zero`
    - Covers the scenario where some validators have Jito rewards and others don’t
    - Ensures validators without Jito entries:
      - Get `jito = 0`
      - Have their `total` computed only from block + inflation rewards

### Testing
I followed the existing CI workflow locally to validate this change:

- Ran the relevant Rust test suite for `validator-debt`, including the new pagination tests
- Ran `bash sh/test_doublezero_solana_clean.sh` 
- Confirmed all tests pass successfully with the updated Jito pagination logic

This gives confidence that:

- The new pagination behavior is covered by unit tests
- Existing behavior for inflation and block rewards remains intact
- Callers of `get_total_rewards` and `get_rewards_between_timestamps` now receive correct Jito rewards even when the API response spans multiple pages

### DIFF
--- a/crates/validator-debt/src/jito.rs
+++ b/crates/validator-debt/src/jito.rs
@@ -1,21 +1,21 @@
 use std::{collections::HashMap, time::Duration};
 
-use anyhow::{Result, anyhow};
+use anyhow::{Context, Result, anyhow};
 use backon::{ExponentialBuilder, Retryable};
 use serde::Deserialize;
-use tracing::info;
+use tracing::{debug, info};
 
 use crate::solana_debt_calculator::ValidatorRewards;
 
 const JITO_BASE_URL: &str = "https://kobe.mainnet.jito.network/api/v1/";
 
+// Safety cap to avoid runaway pagination if the API misbehaves.
+const MAX_JITO_PAGES: u16 = 25;
 pub const JITO_REWARDS_LIMIT: u16 = 1_500;
 
 #[derive(Deserialize, Debug)]
 pub struct JitoRewards {
-    // TODO: check total_count to see if it exceeds entries in a single response
-    // limit - default: 100, max: 10000
-    pub total_count: u16,
+    pub total_count: usize,
     pub rewards: Vec<JitoReward>,
 }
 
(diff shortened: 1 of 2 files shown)


RedaRahmani:
Hey @karl-dz !!
I opened this PR to fix something I noticed in the Jito MEV rewards logic , when the API returns paginated validator_rewards, we were only using the first page, so some rewards could be under-counted for certain epochs.

When you have some time, could you take a look and let me know if the approach feels okay or if you’d prefer it done differently? 
Thanks a lot!

PR #216, opened 2025-11-29
feat(validator-debt): add overrides option to validator debt payment

bgm-malbeclabs:
## Summary of Changes
This PR adds an `overrides.csv` to exclude validators from an epoch or epochs from debt collection.  The attached `overrides.csv` outlines the validators and epochs they are to be excluded from:

[overrides.csv](https://github.com/user-attachments/files/23864374/overrides.csv)

Note that this is a temporary fix; the longer term fix will be using onchain data and writing data onchain.

Closes https://github.com/malbeclabs/doublezero/issues/2316.

## Testing Verification
* This was run locally against epoch 60: 
```
                                                                                                  7VZM7YHcX73TpGoXDeBu61g4QKC86GwAEnew8dA7Y2xn: Some(
                                                                                                                                                           "Program log: Merkle leaf index 89 has already been processed",
                                             )
                                              7ZjHeeYEesmBs4N6aDvCQimKdtJX2bs5boXpJmpG2bZJ: Some(
                                                                                                     "Program log: Insufficient funds in Solana validator deposit to pay debt",
  )
   Validator 7cVfgArCheMR6Cs4t6vz5rfnqd56vZq4ndaBrY5xkxXy for epoch #60 excluded from debt collection
                                                                                                     7knvB4bbqHCKuNp3ef2hJWdwqoH6WAUi55NQt6LdRfkx: Some(
                                                                                                                                                            "Program log: Merkle leaf index 92 has already been processed",
                                              )
```

In the admittedly convoluted logs, the `7cVfgArCheMR6Cs4t6vz5rfnqd56vZq4ndaBrY5xkxXy` is excluded

### DIFF
--- a/crates/validator-debt/src/transaction.rs
+++ b/crates/validator-debt/src/transaction.rs
@@ -282,6 +282,9 @@ impl Transaction {
                 .iter()
                 .any(|(key, epoch)| key == &node_id_str && *epoch == dz_epoch)
             {
+                println!(
+                    "Validator {node_id_str} for epoch #{dz_epoch} excluded from debt collection"
+                );
                 continue;
             }
             let debt_proof = debt_clone.find_debt_proof(&d.node_id).unwrap();


crates/validator-debt/src/transaction.rs
```diff
@@ -256,8 +258,30 @@ impl Transaction {
         let mut debt_collection_result: Vec<DebtCollectionResult> =
             Vec::with_capacity(debt.debts.len());
 
+        let mut overrides = Vec::new();
+
+        if let Ok(file) = File::open("overrides.csv") {
```
bgm-malbeclabs: need to figure out an appropriate path for this

crates/validator-debt/src/transaction.rs
```diff
@@ -256,8 +258,30 @@ impl Transaction {
...
+        }
+
         let debt_clone = debt.clone();
         for d in debt.debts {
+            let node_id_str = d.node_id.to_string();
+            if overrides
+                .iter()
+                .any(|(key, epoch)| key == &node_id_str && *epoch == dz_epoch)
+            {
+                continue;
```
bgm-malbeclabs: maybe add a logline / metric to show what has been skipped

PR #220, opened 2025-12-03
fix(sentinel): improve retry handling for transient RPC errors

vihu:
# Summary

This PR expands the set of retryable error conditions and increases sentinel error retry attempts.

Fixes: https://github.com/malbeclabs/doublezero/issues/2342

## Changes

- Increase max retry attempts from 4 to 10
- Add `is_request()` check to retry hyper `IncompleteMessage` errors
- Add `BrokenPipe` to the list of retryable IO errors
- Treat HTTP 403 as retryable

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2509,6 +2509,7 @@ dependencies = [
  "solana-sdk",
  "solana-system-interface",
  "solana-transaction-status-client-types",
+ "strum",
  "thiserror 2.0.17",
  "tokio",
  "tokio-util",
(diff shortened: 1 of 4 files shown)


snormore (approved):

karl-dz (changes requested):
Actually sorry, should we update the changelog?

bgm-malbeclabs (commented):
since we have prom metrics, it'd be good to record the types of errors we get and their frequency (also changelog)

PR #221, opened 2025-12-03
ci: changelog reminder

snormore:
## Summary of Changes
- Add CI workflow that checks for changelog updates when there are changes to sub-projects that require it, and fails if the changelog wasn't updated
- You can add a `skip-changelog` label to intentionally ignore this check

## Testing Verification
- See PR update history of adding/removing label, and committing test changes


karl-dz (approved):
Finally

PR #237, opened 2025-12-19
validator-debt: enable debt write-off at activation

karl-dz:
## Summary of Changes
This change replaces the flag to enable activation with a check to make sure that the write-offs will occur at the epoch which activates Solana validator debt write-offs.

This PR also:
- Adds the admin command to set this activation epoch.
- Replaces `println` with `tracing` logging in `validator-debt`.
- Performs some clean-up in the `blocks` submodule of `validator-debt`.

Closes https://github.com/malbeclabs/doublezero/issues/2492.

## Testing Verification
Local validator workflow should pass. This test adds an activation epoch, where the first distribution initialization should not trigger the write-off workflow. The subsequent initialization should.

### DIFF
--- a/crates/solana-admin-cli/revenue-distribution/CHANGELOG.md
+++ b/crates/solana-admin-cli/revenue-distribution/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- add `--solana_validator_debt_write_off_feature_activation_epoch` ([#237](https://github.com/doublezerofoundation/doublezero-offchain/pull/237))
 - update migration command ([#229](https://github.com/doublezerofoundation/doublezero-offchain/pull/229))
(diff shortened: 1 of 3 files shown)


crates/validator-debt/src/worker.rs
```diff
@@ -898,33 +896,28 @@ async fn try_write_off_distribution_debt(
...
-            let deposit_balance = deposit_balances.get_mut(&node_id).unwrap();
-
-            let (_, proof) = computed_debt.find_debt_proof(&node_id).unwrap();
-
-            let wallet_key = wallet.pubkey();
+                let deposit_account_info = wallet
+                    .connection
+                    .get_account(&deposit_key)
+                    .await
+                    .unwrap_or_default();
```
vihu: Should we allow this to fail silently or bubble up the error?

vihu (approved):

crates/validator-debt/src/worker.rs
karl-dz: Actually, this would default if the lamports balance is zero. I guess we could disregard trying to initialize the account in this case. We should re-evaluate this in a subsequent PR in case we do not want the accountant to be so charitable

PR #242, opened 2026-01-04
feat(solana-cli): migrate `harvest-2z` Jupiter integration to api.jup.ag + API key auth 

RedaRahmani:
## Related issue:
Closes https://github.com/malbeclabs/doublezero/issues/2509.

## Summary of Changes

- Switch `harvest-2z` Jupiter calls to the new `api.jup.ag` base URL.
- Add `--jupiter-api-key` flag with `JUPITER_API_KEY` env fallback (flag overrides env).
#### Introduce a shared JupiterClient to centralize:
- base URL + `x-api-key` header
- retry policy for transient failures (408/429/5xx) with exponential backoff + jitter
- clearer error messages (status + body snippet; helpful 401 guidance)
- Update `quote.rs` and `swap_instructions.rs` to use the shared client (paths instead of hardcoded endpoints).

## Add targeted unit tests validating:

- api key header is sent

- default base URL is correct

- retries happen on 5xx but not on 400

- missing key returns a clear error

- Security hardening: hide env var values from `--help` output (`hide_env_values = true`) so secrets don’t leak in logs/screenshots.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -667,6 +667,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "assert-json-diff"
+version = "2.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "47e4f2b81832e72834d7518d8487a0396a28cc408186a2e8854c0f98011faf12"
+dependencies = [
+ "serde",
+ "serde_json",
+]
+
+[[package]]
 name = "async-channel"
 version = "1.9.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2234,6 +2244,24 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2a2330da5de22e8a3cb63252ce2abb30116bf5265e89c0e01bc17015ce30a476"
 
 [[package]]
+name = "deadpool"
+version = "0.12.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0be2b1d1d6ec8d846f05e137292d0b89133caf95ef33695424c09568bdd39b1b"
+dependencies = [
+ "deadpool-runtime",
+ "lazy_static",
+ "num_cpus",
+ "tokio",
+]
+
+[[package]]
+name = "deadpool-runtime"
+version = "0.1.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "092966b41edc516079bdf31ec78a2e0588d1d0c08f78b91d8307215928642b2b"
+
+[[package]]
 name = "defer"
 version = "0.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2724,6 +2752,7 @@ dependencies = [
  "tracing",
  "tracing-subscriber",
  "url",
+ "wiremock",
 ]
 
 [[package]]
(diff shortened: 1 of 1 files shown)


crates/solana-cli/src/command/revenue_distribution/harvest_2z/mod.rs
```diff
@@ -25,11 +25,19 @@ const TOKEN_ACCOUNT_RENT_EXEMPTION_LAMPORTS: u64 = 2_039_280;
...
-    /// See https://lite-api.jup.ag/swap/v1/program-id-to-label for available
+    /// See https://api.jup.ag/swap/v1/program-id-to-label for available
     /// program ID labels.
     #[arg(long, value_name = "JUPITER_LABEL")]
     specific_dex: Option<String>,
 
+    #[arg(
+        long,
+        env = "JUPITER_API_KEY",
+        hide_env_values = true,
```
karl-dz: Let's remove environment variable handling

crates/solana-cli/src/command/revenue_distribution/harvest_2z/jupiter/client.rs
```diff
@@ -0,0 +1,334 @@
...
+        let mut hasher = std::hash::DefaultHasher::new();
+        std::thread::current().id().hash(&mut hasher);
+        std::time::Instant::now().hash(&mut hasher);
+        let hash = hasher.finish();
+
+        let jitter_factor = 0.75 + (hash % 50) as f64 / 100.0;
+        let delay_ms = (base_ms as f64 * jitter_factor) as u64;
+
+        tokio::time::sleep(Duration::from_millis(delay_ms)).await;
+    }
```
karl-dz: Can we keep this logic simpler by removing retries? I think most consumers will only have the free tier, so they will be limited to 1 request per second anyway. We can treat the "retryable" errors like any other error, too (so for example, if a consumer is using the same API key from another application, he should be prepared to be faced with a too-many-requests error).

crates/solana-cli/src/command/revenue_distribution/harvest_2z/jupiter/client.rs
```diff
@@ -0,0 +1,334 @@
...
+}
+
+#[allow(dead_code)]
+pub fn load_api_key_from_env() -> Result<String> {
+    std::env::var("JUPITER_API_KEY").map_err(|_| {
+        anyhow::anyhow!("Jupiter API key required. Set JUPITER_API_KEY environment variable")
+    })
+}
```
karl-dz: Can remove this since we do not want to support environment variables

crates/solana-cli/src/command/revenue_distribution/harvest_2z/jupiter/client.rs
```diff
@@ -0,0 +1,334 @@
...
+            .build()
+            .context("Failed to build HTTP client")?;
+
+        Ok(Self { client, base_url })
+    }
+
+    #[allow(dead_code)]
+    pub fn base_url(&self) -> &str {
+        &self.base_url
+    }
```
karl-dz: If nothing is actually using this (since the `JupiterClient` is private within the crate), looks like we can remove this

Cargo.toml
```diff
@@ -40,7 +40,7 @@ md5 = "0.8"
 borsh = { version = "1", features = ["derive"] }
 bytemuck = { version = "1", features = ["derive", "min_const_generics"] }
 chrono = { version = "0.4", features = ["serde"] }
-clap = { version = "4", features = ["derive"] }
+clap = { version = "4", features = ["derive", "env"] }
```
karl-dz: Since we will not support environment variables, we can revert this change

crates/solana-cli/src/command/revenue_distribution/harvest_2z/jupiter/client.rs
```diff
@@ -0,0 +1,334 @@
...
+        body: &impl serde::Serialize,
+    ) -> Result<T> {
+        self.execute_with_retry(|| async {
+            let url = format!("{}{}", self.base_url, path);
+            self.client.post(&url).json(body).send().await
+        })
+        .await
+    }
+
+    async fn execute_with_retry<T, F, Fut>(&self, request_fn: F) -> Result<T>
```
vihu: Please use `backon` crate for retries with exponential backoff, it provides sane defaults and that's the pattern we use pretty much everywhere

RedaRahmani:
Hey @vihu !! thanks for the notes!
I ended up simplifying the client based on the main direction for this PR: removed the retry implementation entirely, so we don’t have constants inside functions / backoff logic here anymore, and kept things focused on URL joining + clear errors.

Also, your notes are definitely in my mind,  if I run into a similar case where retries do make sense, I’ll follow the pattern you suggested (like using `backon`). If you spot anything else in my changes that you’d like adjusted, just let me know and I’ll update it.

RedaRahmani:
Hey @karl-dz , thanks! I just updated the changelog. Everything should be good to go now.

karl-dz (approved):
Looks great, thank you for the contribution @RedaRahmani

PR #247, opened 2026-01-12
feat(validator-debt): parallelize finalize distribution

bgm-malbeclabs:
## Summary of Changes
This PR parallelizes the merkle root check before the distribution is finalized. It uses the same pattern with `stream_iter` as in other places to handle async processing of concurrent requests.

Closes https://github.com/malbeclabs/doublezero/issues/2552.

## Testing Verification
Not relevant (though we should try to add a test for this sometime in the future).


scheduler/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [v0.1.5] - 2026-01-08
 
+- add check and filter for 0 total debt messages posted to slack ([#247](https://github.com/doublezerofoundation/doublezero-offchain/pull/247))
```
karl-dz: We already have a 0.1.6 release. Perhaps we have a 0.1.6 heading with nothing underneath, and then put this under Unreleased?

karl-dz (changes requested):
Looks good. Just one recommendation for the changelog

karl-dz (approved):
Thanks m8

PR #244, opened 2026-01-08
validator-debt: bail early when Revenue Distribution is paused

RedaRahmani:
## Related issue :
Closes to https://github.com/malbeclabs/doublezero/issues/2462.

## Summary of Changes

Added a small pause check at the start of the main `validator-debt` worker entrypoints:

- `finalize_distribution`

- `calculate_distribution`

- `pay_all_solana_validator_debt`

- `try_initialize_distribution`

We fetch the program `ProgramConfig` and check `config.is_paused()`:

- If paused → return Ok(...) early (skip work).

- If not paused → proceed normally.

- Logging (non-spammy)

- `WARN` once per process when we first detect unpaused → paused

- `debug` on subsequent cycles while still paused

- `info` once on paused → unpaused (resume)

## Tests added

Added 3 lightweight tests covering:

- first paused detection returns true (skip) and flips the in-process pause state

- repeated paused checks don’t re-trigger the transition (prevents WARN spam)

- resume (paused → unpaused) flips state back and returns false (proceed)

These tests focus on the transition behavior.

PR #248, opened 2026-01-12
validator-debt: finalize zero debt

karl-dz:
## Summary of Changes
Close the loop on finalizing debt when the calculated debt is zero. Currently, the validator debt process only finalizes if there is computed debt. This change will add the finalize debt instruction to the initialize distribution workflow only if the computed debt for the reward distribution is zero.

Closes https://github.com/malbeclabs/doublezero/issues/2599.

## Testing Verification
CI will pass now.


karl-dz:
This PR will also fix the CI issue in #244 and #247 because the next rewards distribution has zero debt from the mainnet fork

PR #252, opened 2026-01-15
(validator-debt): align Slack summary totals with displayed rows

RedaRahmani:
## Summary 
In the `validator-debt` Slack report, the summary block (“Total Debt Collection”) was aggregating totals across all `DebtCollectionResults`, while the detailed table (“Debt Collected”) was skipping epochs with `total_validators == 0` or `successful_transactions_count == 0`.
That meant the summary ‘Total Debt’ could include epochs not shown in the table, so the total didn’t equal the sum of the visible rows
## What changed
Introduced a small helper module `worker/slack_report.rs` that defines:
- a single visibility predicate for which epochs should show up as Slack rows
- a shared aggregation function to compute summary totals from those same visible rows
Updated:
- post_debt_collection_summary_to_slack to compute totals from visible epochs only
- post_debt_collections_to_slack to render the table using the same visible set
## Tests
Added unit tests for the Slack report aggregation logic, including a regression case with a mix of visible + skipped epochs to ensure:
- summary totals == sum of displayed rows
- skipped epochs do not contribute to totals

Closes : https://github.com/malbeclabs/doublezero/issues/2598

### DIFF
--- a/crates/validator-debt/CHANGELOG.md
+++ b/crates/validator-debt/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Make “Total Debt Collection” Slack summary a global (unfiltered) total while keeping the per-epoch “Debt Collected” table filtered ([#252](https://github.com/doublezerofoundation/doublezero-offchain/pull/252))
 - use vote key from past ([#250](https://github.com/doublezerofoundation/doublezero-offchain/pull/250))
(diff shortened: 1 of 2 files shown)


bgm-malbeclabs:
I like the changes you made with the new module and only showing up when there's actual debt that has been calculated but it would be good to have the `Total Debt Collection` reflect the entirety of all debt collected and outstanding rather than just what's collected during a particular run, I think... 
Screenshot 2026-01-16 at 11 58 05
That was the thought behind the `Total Debt Collection` table and then the rollup were just what was collected during that particular debt collection attempt.

PR #257, opened 2026-01-23
revenue-distribution: add configure-contributor-rewards CLI command

RedaRahmani:
## revenue-distribution: add configure-contributor-rewards CLI command
Closes : https://github.com/malbeclabs/doublezero/issues/2663
## Description

What’s included
- New write CLI command to configure a `ContributorRewards` account:
- Accepts repeated `--recipient PUBKEY:PERCENT ` inputs (human percent 1–100) and converts internally to basis points.

Supports mutually exclusive flags:
- `--block-protocol-management`
- `--allow-protocol-management`
- `Block/allow` flag instruction is only sent when requested, If you don’t pass `--block-protocol-management` or `--allow-protocol-management` , the command does not change the “protocol management blocked” setting. It will only update recipients (if you provided `--recipient`). If you pass block or allow, it includes that extra `on-chain` instruction.

Strong input validation:
- Rejects malformed `PUBKEY:PERCENT`, invalid pubkeys, duplicates, out-of-range percents, and totals that don’t sum to 100.
- Errors on “nothing to do” (no recipients + no block/allow).
- Compute budget + optional priority fee, built via a testable helper , The transaction includes a `ComputeBudget` CU limit instruction so the `tx` has enough compute. If you set `--with-compute-unit-price`, it also includes a priority fee (CU price) instruction. Instruction building is done in a small helper function so we can unit test the exact instruction list (order + what’s included) without RPC/sending a tx.
- Preflight check: fetches `ContributorRewards` and fails early if `signer != current rewards_manager_key`.

### Tests
- Added unit tests covering:
  - `PUBKEY:PERCENT` parsing (missing/extra `:`, invalid pubkey, non-numeric percent)
  - percent range validation (reject 0 and >100) + total must sum to 100
  - duplicate recipients + max recipients limit
  - mutual exclusivity of `--block-protocol-management` / `--allow-protocol-management`
  - instruction planning behavior
  
  Validated end-to-end on localnet.

### DIFF
--- a/crates/solana-cli/CHANGELOG.md
+++ b/crates/solana-cli/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+- solana-cli: add `revenue-distribution configure-contributor-rewards` command to update ContributorRewards recipients and (optionally) protocol-management block/allow flags ([#257](https://github.com/doublezerofoundation/doublezero-offchain/pull/257))
 - ensure debt is finalized before collection ([#268](https://github.com/doublezerofoundation/doublezero-offchain/pull/268))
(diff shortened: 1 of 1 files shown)


RedaRahmani:
Hey @karl-dz , Thanks for the update and the review!  I’m happy to tweak anything around the CLI UX behavior if you’d like changes, just let me know.

PR #265, opened 2026-02-09
feat: billing sentinel for tenant payment status monitoring

vihu:
# Summary

  Add a billing sentinel that periodically checks tenant token accounts, deducts epoch fees via SPL transfers, and updates payment status on the DZ Ledger. Handles both the
  new flat-per-epoch deduction model (rate > 0) and the legacy balance-check path (rate == 0).

## Changes

  - Add BillingSentinel with configurable poll interval, minimum balance threshold, and persistent state directory
  - Deduct tenants one epoch at a time via transfer_checked to a journal ATA, then bump the billing epoch on the DZ Ledger
  - Mark tenants delinquent when their token account balance falls below the required rate
  - Verify token account ownership (sentinel authority) before reading balances to catch misconfigured accounts early
  - Use a two-phase deduction cache (Transferred / Completed / Failed) to prevent double-charges when the epoch update fails after a successful transfer
  - Persist pending-transfer markers to disk so deduction state survives process restarts
  - Add billing_state_dir to sentinel settings (default ./billing-state, configurable via config file or SENTINEL__BILLING_STATE_DIR)
  - Add billing_poll_interval and minimum_balance CLI args
  - Extend SolRpcClientType trait with get_token_account_balance and transfer_spl_token
  - Extend DzRpcClientType trait with update_tenant_payment_status, update_tenant_billing_epoch, and get_current_dz_epoch

### DIFF
--- a/crates/sentinel/src/settings.rs
+++ b/crates/sentinel/src/settings.rs
@@ -126,10 +126,11 @@ impl Settings {
 
     pub fn doublezero_mint(&self) -> Pubkey {
         match self.env.to_lowercase().as_str() {
-            "mainnet" | "mainnet-beta" => {
+            "mainnet" | "mainnet-beta" | "local" | "localhost" => {
+                // NOTE: local|localhost are forked off of mn-beta
                 doublezero_revenue_distribution::env::mainnet::DOUBLEZERO_MINT_KEY
             }
-            "testnet" | "devnet" | "local" | "localhost" => {
+            "testnet" | "devnet" => {
                 doublezero_revenue_distribution::env::development::DOUBLEZERO_MINT_KEY
             }
             other => {


elitegreg (dismissed):

crates/sentinel/src/client/solana.rs
```diff
@@ -332,6 +364,61 @@ impl SolRpcClient {
...
+            to,
+            &signer.pubkey(),
+            &[],
+            amount,
+            decimals,
+        )
+        .map_err(|e| Error::SplInstruction(format!("transfer_checked: {e}")))?;
+
+        let compute_limit_ix = ComputeBudgetInstruction::set_compute_unit_limit(12_000);
+        let compute_price_ix = ComputeBudgetInstruction::set_compute_unit_price(100_000);
```
karl-dz: This also might not be necessary. Currently the debt (and I think the rewards) accountants do not specify CU prices and they land transactions just fine.

You may want to have this configurable if you wanted to keep it, though

Issue #274, opened 2026-03-02
Reservation CLI: SDK, CLI, and mock program


martinsander00:
Split implementation of the reservation CLI tooling.

## Tasks
- [x] **SDK**: instruction builders, PDA helpers, raw-byte account parsers (DeviceHistory, MetroHistory, ClientSeat, PaymentEscrow)
- [x] **CLI**: `reservation` subcommand (`initialize-seat`, `withdraw`, `list`, `price`) behind `experimental` feature gate

## Context
- `price` command fetches MetroHistory + DeviceHistory via getProgramAccounts, joins client-side, displays base price + device premium
- All commands are behind the `experimental` feature gate
- Mock program mirrors the real on-chain program's account layouts for realistic integration testing

PR #276, opened 2026-03-02
reservation: add CLI commands (initialize-seat, withdraw, list, price)

martinsander00:
## Summary of Changes
- Add `reservation` subcommand behind the `experimental` feature gate with:
  - `initialize-seat`: create a client seat and payment escrow for a (device, client_ip) pair; skips seat creation if it already exists
  - `fund`: deposit USDC into a per-seat payment escrow; validates positive amount and checks on-chain minimum price
  - `withdraw`: close a payment escrow and reclaim USDC; pre-flight check for escrow existence
  - `list`: show all client seats with optional device, client-ip, and withdraw-authority filters
  - `price`: show current device pricing (metro base + device premium)
- Simplify `ClosePaymentEscrowAccounts` SDK constructor to accept `Option<&Pubkey>` for refund account

## Diff Breakdown
| Category          | Files | Lines (+/-) | Net   |
|-------------------|-------|-------------|-------|
| CLI commands      |     7 | +798 / -0   | +798  |
| SDK (constructor) |     1 | +10 / -13   | -3    |
| Build / changelog |     2 | +5 / -0     | +5    |

~800 net lines — 5 new CLI commands, subcommand scaffold, and SDK cleanup.

<details>
<summary>Key files (click to expand)</summary>

- `crates/solana-cli/src/command/reservation/mod.rs` — subcommand enum, `DeviceArgs` (shared `--device` / `--device-code` resolution), feature gate
- `crates/solana-cli/src/command/reservation/initialize_seat.rs` — sends `InitializeClientSeat` + `InitializePaymentEscrow` in one transaction; skips seat ix if already exists
- `crates/solana-cli/src/command/reservation/fund.rs` — sends `FundPaymentEscrowUsdc`, derives exchange key from on-chain DeviceHistory, validates amount against current price
- `crates/solana-cli/src/command/reservation/withdraw.rs` — sends `ClosePaymentEscrow`, pre-flight escrow existence check
- `crates/solana-cli/src/command/reservation/price.rs` — read-only, joins MetroHistory + DeviceHistory to show per-device pricing
- `crates/solana-cli/src/command/reservation/list.rs` — read-only, fetches ClientSeat accounts with discriminator memcmp filter, supports `--withdraw-authority` and `--client-ip` filters
- `crates/solana-cli/src/command/mod.rs` — wires `Reservation` variant into top-level CLI
- `crates/solana-sdk/src/reservation/instruction/account.rs` — merged `new` / `new_with_ata_refund` into single constructor

</details>

## Related PRs
- **#275 (SDK)**: merged — this PR builds on it
- **#277 (Mock + tests)**: depends on this — adds mock program and integration tests

Part 2 of #274

### DIFF
--- a/crates/solana-cli/src/command/reservation/price.rs
+++ b/crates/solana-cli/src/command/reservation/price.rs
@@ -19,11 +19,11 @@ use tabled::{Table, Tabled, settings::Style};
 #[derive(Debug, Args)]
 pub struct PriceCommand {
     /// Filter by device public key.
-    #[arg(long)]
+    #[arg(long, group = "filter")]
     device: Option<Pubkey>,
 
     /// Filter by metro exchange public key.
-    #[arg(long)]
+    #[arg(long, group = "filter")]
     metro: Option<Pubkey>,
 
     #[command(flatten)]


crates/solana-cli/src/command/reservation/fund.rs
```diff
@@ -0,0 +1,114 @@
...
+
+        let device = self.device_args.resolve(network_env).await?;
+        let usdc_mint_key = self.usdc_mint.unwrap_or(*state::USDC_MINT_KEY);
+        let client_ip_bits = u32::from(self.client_ip);
+
+        // Convert decimal USDC to micro-USDC (6 decimals).
+        let amount_micro = (self.amount * 1_000_000.0).round() as u64;
+        if amount_micro == 0 {
+            bail!("Amount must be greater than zero");
+        }
```
karl-dz: Couple things:
1. What happens if f64 is a negative value? Can we revert if it isn't positive?
2. Should we fetch the device history to try to get the current price? A revert will say in the program logs that the amount must be X, but I'm not sure that will be good UX. We may want to have a clean client error saying that the amount needs to be at least X.
  - Looks like you fetch the device history account below, so it should be easy to check the current price.

martinsander00: A negative f64 cast to u64 in Rust is saturating so it produces 0 (not a garbage/wrapping value as in C). So the old code would have caught it with the amount_micro == 0 check, but the error message "Amount must be greater than zero" was pretty bad

Moved the validation before the conversion

PR #341, opened 2026-04-07
shreds pay: check request pda to close race in duplicate ip guard

packethog:
## Summary

- Close a race window in the duplicate IP guard from #340. Two concurrent `pay` calls for the same IP on different devices could both pass the tenure check (both seats have tenure=0 before the oracle processes them).
- For seats on other devices with tenure=0, check for an `InstantSeatAllocationRequest` PDA. If present, the seat is in-flight (just paid, waiting for oracle) and blocks. If absent, the seat was withdrawn and is safe to ignore.

## Testing Verification

- 87 unit tests passing (added `withdrawn_seat_goes_to_pending` and `mixed_active_and_pending` tests)
- Clippy and fmt clean

### DIFF
--- a/crates/solana-cli/CHANGELOG.md
+++ b/crates/solana-cli/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 - `shreds pay`: block duplicate client IP across devices — prevent creating a seat for an IP that already has an active seat on a different device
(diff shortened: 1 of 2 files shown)


packethog:
Closing — the request PDA check is a no-op. tenure_epochs is set to 1 in the same atomic instruction that creates the request PDA, so there is no observable state where tenure=0 and the request PDA exists. The tenure>0 check from #340 already covers all catchable cases. The concurrent race (two pay txs in flight simultaneously) cannot be fixed client-side.

PR #347, opened 2026-04-14
solana-cli: shreds withdraw supports stale seats and --funds-only

martinsander00:
## Summary
- Previously `shreds withdraw` required the seat to have active service, leaving users with stale seats unable to recover USDC from their payment escrow
- The CLI now auto-detects stale seats and skips `RequestInstantSeatWithdrawal`, closing only the payment escrow to return funds
- Adds `--funds-only` flag to explicitly skip the instant withdrawal request even when the seat is active

## Lines of Code
| Section | Added | Removed |
|---------|-------|---------|
| SDKs | +25 | -10 |

### DIFF
--- a/crates/solana-cli/src/command/shreds/withdraw.rs
+++ b/crates/solana-cli/src/command/shreds/withdraw.rs
@@ -99,9 +99,9 @@ impl WithdrawCommand {
         // (or --funds-only) skip the request and just close the payment escrow.
         let request_instant_withdrawal = has_active_service && !self.funds_only;
 
-        if !has_active_service && !escrow_exists {
+        if !request_instant_withdrawal && !escrow_exists {
             bail!(
-                "Client seat {client_seat_key} does not have active service and no payment escrow exists"
+                "Client seat {client_seat_key} has no payment escrow to close and no active service to withdraw"
             );
         }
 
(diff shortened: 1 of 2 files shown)


bgm-malbeclabs:
logic looks sound but can we add a test?

martinsander00:
> logic looks sound but can we add a test?

yessir, working on it

crates/solana-cli/src/command/shreds/withdraw.rs
```diff
@@ -82,27 +86,36 @@ impl WithdrawCommand {
...
         let current_epoch = wallet.connection.get_epoch_info().await?.epoch;
-        if active_epoch < current_epoch {
-            bail!("Client seat {client_seat_key} does not have active service");
+        let has_active_service = active_epoch >= current_epoch;
+
+        // Only request instant withdrawal when the seat is active. Stale seats
+        // (or --funds-only) skip the request and just close the payment escrow.
+        let request_instant_withdrawal = has_active_service && !self.funds_only;
+
+        if !has_active_service && !escrow_exists {
```
karl-dz: Shouldn't this just be `!escrow_exists`? I feel like this would short-circuit if the user has active service, so we would pass through to the end with no instructions besides the compute budget ones. Can you double-check me?

martinsander00:
> logic looks sound but can we add a test?


will add e2e in shreds repo

karl-dz (approved):

PR #350, opened 2026-04-20
solana-cli: shreds pay integrates prorated instant seat allocation

martinsander00:
Resolves: malbeclabs/infra#1062

## Summary
- Teaches `shreds pay` about `ProgramConfig.is_prorated_service_enabled` (bit 2 of the flags word, see malbeclabs/doublezero-shreds#258) so users can pay late in the epoch without being blocked.
- When the bit is set: skip the client-side min-price check and suppress the late-epoch warning — the onchain program charges the prorated amount and enforces its own floor.
- When the bit is unset: preserve legacy full-epoch behavior. Pre-prorated programs naturally read `false` because the bit sits at a stable offset right after the discriminator.
- Folds the `ProgramConfig` fetch into the existing `get_multiple_accounts` call so there is no extra RPC.

## Lines of Code
| Section | Added | Removed |
|---------|-------|---------|
| Core logic | +30 | -18 |
| SDKs | +38 | -0 |
| Tests | +67 | -3 |

## Testing Verification
- New `no_warning_when_prorated_service_enabled` test in `pay.rs` covers the warning-suppression branch.
- New SDK test module in `crates/solana-sdk/src/shred_subscription/state.rs` covers `is_prorated_service_enabled` against bit set, bit unset, unrelated bits, short buffers, and empty buffers.

### DIFF
--- a/crates/solana-cli/CHANGELOG.md
+++ b/crates/solana-cli/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- `shreds pay`: integrate prorated instant seat allocation — when the onchain `is_prorated_service_enabled` flag is set, skip the client-side min-price check and suppress the late-epoch warning; legacy behavior preserved when the flag is unset
+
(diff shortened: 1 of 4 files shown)


bgm-malbeclabs (approved):
logic looks sound

PR #360, opened 2026-05-08
feat(solana-cli): add validator rewards commands

bgm-malbeclabs:
## Summary of Changes
This PR (bgm/configure_dst) introduces a new subcommand group on doublezero-solana shreds for managing the per-validator ValidatorPublisherRewards PDA — the on-chain account that controls where a validator's shred-publisher rewards are paid (the destination token mint and the wallet that owns the receiving ATA).

An ATA is created if it doesn't exist; idempotent if it does. On `show` it will show `ATA status: exists` or `ATA status: missing — re-run configure to create it`. 

## Testing Verification
* This has been manually tested
* Solana fork has been updated to exercise the new commands


crates/solana-cli/src/command/shreds/publisher_rewards/configure.rs
```diff
@@ -0,0 +1,431 @@
...
+    pub rewards_token_mint: Pubkey,
+
+    /// Wallet that will own the ATA that receives rewards.
+    #[arg(long)]
+    pub rewards_token_owner: Pubkey,
+
+    /// Path to the validator identity keypair file. Direct path: this signs
+    /// the transaction. Mutually exclusive with `--signature`.
+    #[arg(long, conflicts_with_all = ["signature", "deadline_slot"])]
+    pub validator_identity_keypair: Option<String>,
```
karl-dz: `--keypair` / `-k` should already cover this, right? It seems like we should lean on the optional signature argument. It shouldn't matter whether the validator identity is the keypair used to execute the transaction if a signature is provided, if that makes sense

crates/solana-cli/src/command/shreds/publisher_rewards/prepare_offchain_message.rs
```diff
@@ -0,0 +1,186 @@
...
+    if deadline_slot.is_some() && valid_for.is_some() {
+        // Defense in depth — clap also catches this via `conflicts_with`.
+        return Err(anyhow!(
+            "--deadline-slot and --valid-for are mutually exclusive"
+        ));
+    }
+    if let Some(d) = deadline_slot {
+        return Ok(d);
+    }
+    let duration = valid_for.unwrap_or(Duration::from_secs(60 * 60));
```
karl-dz: Can this default of 1 hour be defined as a const?

crates/solana-cli/src/command/shreds/publisher_rewards/show.rs
```diff
@@ -0,0 +1,63 @@
...
+                    "Validator publisher rewards not initialized for {}",
+                    self.node_id
+                );
+                Ok(())
+            }
+            Some(account) => {
+                let (node_id, owner, mint) = parse_validator_publisher_rewards(&account.data)
+                    .ok_or_else(|| {
+                        anyhow::anyhow!("VPR account at {pda} is malformed (data too short)")
+                    })?;
```
karl-dz: Can we find all instances of `ok_or_else` and replace these with `context` or `with_context`?

crates/solana-cli/src/command/shreds/publisher_rewards/show.rs
```diff
@@ -0,0 +1,63 @@
...
+    pub async fn try_into_execute(self) -> Result<()> {
+        let connection = self.connection_options.into_shred_subscription_connection();
+        let pda = find_validator_publisher_rewards_address(&self.node_id).0;
+
+        // `get_account_with_commitment` returns `Ok(Response { value: None })`
+        // for missing accounts and `Err(...)` for transport errors. We only
+        // print "not initialized" on the former; transport errors propagate
+        // so an RPC outage doesn't silently look like a missing PDA.
+        let account = connection
+            .get_account_with_commitment(&pda, CommitmentConfig::confirmed())
```
karl-dz: I thought the client defaults to confirmed commitment

crates/solana-cli/src/command/shreds/publisher_rewards/show.rs
```diff
@@ -0,0 +1,63 @@
...
+                println!(
+                    "Validator publisher rewards not initialized for {}",
+                    self.node_id
+                );
+                Ok(())
+            }
+            Some(account) => {
+                let (node_id, owner, mint) = parse_validator_publisher_rewards(&account.data)
+                    .ok_or_else(|| {
+                        anyhow::anyhow!("VPR account at {pda} is malformed (data too short)")
```
karl-dz: Can we avoid using abbreviations? VPR might not be clear to the user

crates/solana-cli/src/command/shreds/publisher_rewards/show.rs
karl-dz: Also, we should use `get_account` anyway because the option is a bit awkward.

https://docs.rs/solana-client/latest/solana_client/nonblocking/rpc_client/struct.RpcClient.html#method.get_account

crates/solana-cli/src/command/shreds/publisher_rewards/show.rs
```diff
@@ -0,0 +1,63 @@
...
+            .context("failed to fetch validator publisher rewards account")?
+            .value;
+
+        match account {
+            None => {
+                println!(
+                    "Validator publisher rewards not initialized for {}",
+                    self.node_id
+                );
+                Ok(())
```
karl-dz: Perhaps this should be an error (and it will be once you change the above to `get_account`)

crates/solana-cli/src/command/shreds/publisher_rewards/show.rs
karl-dz: Also I believe we can just use `ZeroCopyAccountOwned::<ValidatorPublisherRewards>::try_from` instead here. We don't need to reinvent the wheel with `parse_validator_publisher_rewards` which does some weird deserialization

crates/solana-cli/src/command/shreds/publisher_rewards/show.rs
```diff
@@ -0,0 +1,63 @@
...
+            }
+            Some(account) => {
+                let (node_id, owner, mint) = parse_validator_publisher_rewards(&account.data)
+                    .ok_or_else(|| {
+                        anyhow::anyhow!("VPR account at {pda} is malformed (data too short)")
+                    })?;
+                let ata = get_associated_token_address(&owner, &mint);
```
karl-dz: We should actually try to fetch the ATA and error out if it doesn't exist. Rewards won't be distributed to validators if there is no ATA existing with their elected token owner and preferred mint

crates/solana-cli/src/command/shreds/publisher_rewards/configure.rs
```diff
@@ -0,0 +1,386 @@
...
+pub struct ConfigureCommand {
+    /// Validator node identity. Required in the offchain path; defaults to
+    /// the fee-payer keypair's pubkey in the direct path.
+    #[arg(long)]
+    pub node_id: Option<Pubkey>,
+
+    /// Mint to receive rewards in. Must correspond to a registered, enabled
+    /// `ShredRewardToken`.
+    #[arg(long)]
+    pub rewards_token_mint: Pubkey,
```
karl-dz: Let's make this optional. It can default to 2Z

karl-dz: Also, perhaps we introduce a mint alias instead of having to provide a pubkey ("2z", "usdc", "wsol")

crates/solana-cli/src/command/shreds/publisher_rewards/configure.rs
```diff
@@ -0,0 +1,386 @@
...
+       --rewards-token-mint <MINT> --rewards-token-owner <WALLET> \
+       --signature <BASE58> --deadline-slot <ABS>
+*/
+
+#[derive(Debug, Args)]
+pub struct ConfigureCommand {
+    /// Validator node identity. Required in the offchain path; defaults to
+    /// the fee-payer keypair's pubkey in the direct path.
+    #[arg(long)]
+    pub node_id: Option<Pubkey>,
```
karl-dz: Maybe this should be required for any interaction, just so the user is explicitly specifying which validator identity is being configured.

cc: @martinsander00 what are your thoughts about this?

crates/solana-cli/src/command/shreds/publisher_rewards/configure.rs
```diff
@@ -0,0 +1,386 @@
...
+                self.rewards_token_mint
+            );
+        }
+        let vpr_exists = accounts.get(1).and_then(|a| a.as_ref()).is_some();
+
+        // Build instructions.
+        let mut instructions = vec![super::super::build_check_cli_version_instruction()?];
+
+        if !vpr_exists {
+            println!("VPR PDA missing; will initialize as part of this transaction.");
```
karl-dz: Another abbreviation. We should go through the changes in this branch to make sure we don't abbreviate any user-facing messages

crates/solana-cli/src/command/shreds/publisher_rewards/prepare_offchain_message.rs
```diff
@@ -0,0 +1,187 @@
...
+const DEFAULT_VALID_FOR: Duration = Duration::from_secs(60 * 60);
+
+#[derive(Debug, Args)]
+pub struct PrepareOffchainMessageCommand {
+    #[arg(long)]
+    pub node_id: Pubkey,
+    #[arg(long)]
+    pub rewards_token_mint: Pubkey,
+    #[arg(long)]
+    pub rewards_token_owner: Pubkey,
```
karl-dz, line 33: We need to make sure this pubkey is not the default pubkey

crates/solana-cli/src/command/shreds/publisher_rewards/configure.rs
```diff
@@ -0,0 +1,386 @@
...
+    pub node_id: Option<Pubkey>,
+
+    /// Mint to receive rewards in. Must correspond to a registered, enabled
+    /// `ShredRewardToken`.
+    #[arg(long)]
+    pub rewards_token_mint: Pubkey,
+
+    /// Wallet that will own the ATA that receives rewards.
+    #[arg(long)]
+    pub rewards_token_owner: Pubkey,
```
karl-dz, line 62: We need to make sure this pubkey is not the default pubkey

crates/solana-cli/src/command/shreds/publisher_rewards/configure.rs
```diff
@@ -0,0 +1,386 @@
...
+                    "in the direct path the fee-payer keypair signs as the validator identity, \
+                     but its pubkey {fee_payer_pubkey} does not match --node-id {node_id}. \
+                     Either drop --node-id, pass a fee-payer keypair (-k) whose pubkey is {node_id}, \
+                     or use the offchain path with --signature + --deadline-slot."
+                );
+            }
+            Ok((node_id, ResolvedAuth::Direct))
+        }
+        // clap's `requires` ensures signature and deadline_slot are both-or-neither.
+        _ => bail!("internal error: clap should have rejected this combination"),
```
karl-dz: This feels a bit awkward. Shouldn't the deadline slot and signature be passed in as an optional tuple instead? When passing this in, you can `zip` two options.

For reference: https://doc.rust-lang.org/std/option/enum.Option.html#method.zip

crates/solana-cli/src/command/shreds/publisher_rewards/configure.rs
```diff
@@ -0,0 +1,386 @@
...
+        let accounts = wallet
+            .connection
+            .get_multiple_accounts(&[srt_pda, vpr_pda])
+            .await
+            .context("failed to read pre-flight accounts")?;
+
+        let srt_account = accounts.first().and_then(|a| a.as_ref()).ok_or_else(|| {
+            anyhow!(
+                "rewards token mint {} is not a registered ShredRewardToken — \
+                 ask an admin to register it via InitializeShredRewardToken",
```
karl-dz: We can remove "ask an admin" because that will never happen

crates/solana-sdk/src/shred_subscription/state.rs
```diff
@@ -502,6 +518,87 @@ pub fn parse_metro_history(data: &[u8]) -> Option<MetroHistoryInfo> {
...
+    );
+    let flags = u64::from_le_bytes(
+        data[SHRED_REWARD_TOKEN_FLAGS_OFFSET..SHRED_REWARD_TOKEN_FLAGS_OFFSET + 8]
+            .try_into()
+            .ok()?,
+    );
+    let is_enabled = flags & SHRED_REWARD_TOKEN_FLAG_IS_ENABLED_BIT != 0;
+    Some((mint, is_enabled))
+}
+
```
karl-dz, line 685: We should define the accounts as they are defined in the other repo instead of having these deserializers

crates/solana-sdk/src/shred_subscription/state.rs
```diff
@@ -570,4 +667,83 @@ mod tests {
...
+    }
+
+    #[test]
+    fn shred_reward_token_seeds_and_address_are_deterministic() {
+        let mint = Pubkey::new_unique();
+        let (a1, b1) = find_shred_reward_token_address(&mint);
+        let (a2, b2) = find_shred_reward_token_address(&mint);
+        assert_eq!(a1, a2);
+        assert_eq!(b1, b2);
+    }
```
karl-dz: These tests exercise `find_program_address` and are not needed

crates/solana-cli/src/command/shreds/publisher_rewards/show.rs
karl-dz: Same with `map_err`. Anything where these methods use the `anyhow!` macro

crates/solana-cli/src/command/shreds/publisher_rewards/configure.rs
bgm-malbeclabs: feels like it should be required yea

martinsander00 (dismissed):

karl-dz (approved):

PR #365, opened 2026-05-11
feat(solana-cli): allow validator client teams to claim rewards (#1201)

bgm-malbeclabs:
## Summary of Changes
This PR adds in commands for validators to claim their rewards for publishing shreds in an approved currency. 

Gist outlining new [commands](https://gist.github.com/bgm-malbeclabs/28f0e9a73494c974c5fd85dcfce854a4)

## Testing Verification
* Unit tests have been added to exercise new commands
* Solana local fork has been updated to exercise new commands as well

### DIFF
--- a/.github/workflows/local-validator.yml
+++ b/.github/workflows/local-validator.yml
@@ -34,8 +34,12 @@ jobs:
           echo "$HOME/.local/share/solana/install/active_release/bin" >> $GITHUB_PATH
       - name: Generate ~/.config/solana/id.json
         run: solana-keygen new --silent --no-bip39-passphrase
+      - name: Generate manager keypair for synthetic VCR
+        run: solana-keygen new --silent --no-bip39-passphrase -o manager_keypair.json
       - name: Start Solana mainnet-beta fork in background
-        run: cargo run --bin doublezero-solana-fork -- -um --reset > /dev/null 2>&1 &
+        run: |
+          MANAGER_PUBKEY=$(solana address -k manager_keypair.json)
+          cargo run --bin doublezero-solana-fork -- -um --reset --synthetic-vcr-manager "$MANAGER_PUBKEY" > /dev/null 2>&1 &
       - name: Build `doublezero-solana`
         run: cargo build --bin doublezero-solana
       - name: Run `doublezero-solana` tests
--- a/.github/workflows/local-validator.yml
+++ b/.github/workflows/local-validator.yml
@@ -34,7 +34,7 @@ jobs:
           echo "$HOME/.local/share/solana/install/active_release/bin" >> $GITHUB_PATH
       - name: Generate ~/.config/solana/id.json
         run: solana-keygen new --silent --no-bip39-passphrase
-      - name: Generate manager keypair for synthetic VCR
+      - name: Generate manager keypair for synthetic ValidatorClientRewards
         run: solana-keygen new --silent --no-bip39-passphrase -o manager_keypair.json
       - name: Start Solana mainnet-beta fork in background
         run: |
--- a/crates/solana-cli/src/command/shreds/validator_client_rewards/claim.rs
+++ b/crates/solana-cli/src/command/shreds/validator_client_rewards/claim.rs
@@ -64,8 +64,23 @@ pub(crate) fn validate_manager(wallet: &Pubkey, vcr_manager: &Pubkey) -> Result<
     Ok(())
 }
 
+/// Upper bound on epochs per claim tx. Each `ClaimHoldingId` adds 9 bytes of
+/// instruction data and the holding account adds 32 bytes to the account list,
+/// so beyond ~20 the tx blows past the 1232-byte packet limit. 16 is a
+/// conservative cap that leaves room for the destination/rent/program-config
+/// accounts and the CheckCliVersion ix.
+pub(crate) const MAX_CLAIM_EPOCHS_PER_TX: usize = 16;
+
 impl ClaimCommand {
     pub async fn try_into_execute(self) -> Result<()> {
+        if self.subscription_epochs.len() > MAX_CLAIM_EPOCHS_PER_TX {
+            bail!(
+                "too many --subscription-epoch values ({}); max {} per tx. Split into multiple `claim` calls.",
+                self.subscription_epochs.len(),
+                MAX_CLAIM_EPOCHS_PER_TX
+            );
+        }
+
         let dz_connection = self
             .solana_payer_options
             .connection_options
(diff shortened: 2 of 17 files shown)


.github/workflows/local-validator.yml
```diff
@@ -34,8 +34,12 @@ jobs:
           echo "$HOME/.local/share/solana/install/active_release/bin" >> $GITHUB_PATH
       - name: Generate ~/.config/solana/id.json
         run: solana-keygen new --silent --no-bip39-passphrase
+      - name: Generate manager keypair for synthetic VCR
```
martinsander00: VCR -> ValidatorClientRewards :)

crates/solana-cli/src/command/shreds/validator_client_rewards/claim.rs
```diff
@@ -0,0 +1,403 @@
...
+                missing
+            );
+        }
+        if !wrong_owner.is_empty() {
+            bail!(
+                "claim holdings for epochs {:?} are not SPL token accounts (owners: {:?})",
+                wrong_owner.iter().map(|(e, _)| *e).collect::<Vec<_>>(),
+                wrong_owner
+            );
+        }
```
martinsander00: if both `wrong_owner` and `wrong_mint` are non-empty the user only sees `wrong_owner` and has to fix-retry. can we combine these into one diagnostic?

martinsander00 (commented):
i see a few places where the type is explicitly annotated when the compiler could infer it. can we drop them? e.g. in claim.rs:

```rust
let mut missing: Vec<u64> = Vec::new();
let mut wrong_owner: Vec<(u64, Pubkey)> = Vec::new();
let mut total_drained: u64 = 0;
let mut pre_claim_balances: Vec<(u64, Pubkey, u64)> = Vec::new();
let claim_holding_ids: Vec<ClaimHoldingId> = ...
let metas: Vec<AccountMeta> = claim_accounts.into();
let cu_limit: u32 = 30_000u32.saturating_mul(...);
```

also the `30_000u32` / `25_000u32` u32s can be dropped

and on the anyhow side, a few `.ok_or_else(|| anyhow!(...))` calls on options could just be `.context(...)` / `.with_context(...)` (claim.rs:96, 104, 110, 112, 190).

crates/solana-cli/src/command/shreds/validator_client_rewards/claim.rs
```diff
@@ -0,0 +1,403 @@
...
+            &ID,
+            metas,
+            &ShredSubscriptionInstructionData::ClaimValidatorClientRewards(claim_holding_ids),
+        )?;
+
+        let mut instructions = vec![super::super::build_check_cli_version_instruction()?, ix];
+
+        // ~30k CU per holding (token transfer + close + state decrement), plus
+        // the check-cli-version ix.
+        let cu_limit: u32 = 30_000u32.saturating_mul(self.subscription_epochs.len() as u32 + 1);
```
martinsander00, line 277: no upper bound on `subscription_epochs` right? if someone passes too many we'll just get an rpc error when the tx goes above 1232 bytes. can we cap it or fail fast?

crates/solana-cli/src/command/shreds/validator_client_rewards/claim.rs
```diff
@@ -0,0 +1,403 @@
...
+        let tx_outcome = wallet.send_or_simulate_transaction(&transaction).await?;
+
+        if let TransactionOutcome::Executed(tx_sig) = tx_outcome {
+            println!("Claimed: {tx_sig}");
+            // The on-chain handler transfers the full balance of each holding,
+            // so the pre-claim balance equals what was drained.
+            for (epoch, holding_pda, drained) in &pre_claim_balances {
+                println!("  epoch {epoch}: {drained} from {holding_pda}");
+            }
+            println!("Total claimed: {total_drained}");
```
martinsander00: this is the pre-claim balance, not what was actually drained right? if the holding gets topped up between the `getMultipleAccounts` and the tx landing, the printed `Total claimed` is stale. maybe rename to "pre-claim balance" or pull the destination diff post-tx?

crates/solana-cli/src/command/shreds/validator_client_rewards/show.rs
```diff
@@ -0,0 +1,252 @@
...
+                .await
+                .with_context(|| format!("fetching manager ATA {manager_ata}"))?
+                .value;
+            let ata_amount = match ata_account {
+                Some(acct) if acct.owner == spl_token_interface::ID => {
+                    spl_token_interface::state::Account::unpack(&acct.data)
+                        .ok()
+                        .map(|t| t.amount)
+                }
+                _ => None,
```
martinsander00: this swallows `wrong-owner / unpack-fail / doesn't-exist` all as `(not initialized)` right? `claim.rs` splits these cases, can we do the same here so a user with a non-token account at the ATA doesn't get confused?

crates/solana-cli/src/command/shreds/validator_client_rewards/show.rs
```diff
@@ -0,0 +1,252 @@
...
+                    .zip(holding_keys.iter())
+                    .zip(holding_accounts.into_iter())
+                {
+                    let amount = match maybe_acct {
+                        Some(acct) if acct.owner == spl_token_interface::ID => {
+                            spl_token_interface::state::Account::unpack(&acct.data)
+                                .ok()
+                                .map(|t| t.amount)
+                        }
+                        _ => None,
```
martinsander00: same here

crates/solana-cli/src/command/shreds/validator_client_rewards/init_holding.rs
```diff
@@ -0,0 +1,231 @@
...
+                    &wallet_key,
+                ),
+                &ShredSubscriptionInstructionData::InitializeClaimHoldingAccount(*epoch),
+            )?;
+            instructions.push(ix);
+        }
+
+        // Allow ~25k CU per init (system create + spl-token init + state update),
+        // plus one for the check-cli-version ix.
+        let cu_limit: u32 = 25_000u32.saturating_mul(to_init.len() as u32 + 1);
```
martinsander00, line 150: same point as on `claim.rs`, no upper bound on `subscription_epochs` count, large N will fail with a tx-too-big error

crates/solana-fork/src/main.rs
```diff
@@ -621,3 +663,63 @@ fn try_dump_program(
...
+    };
+    use solana_sdk::rent::Rent;
+
+    // Account size = discriminator (8) + struct body. The struct body is 176
+    // bytes: header (8: client_id + bump + padding) + manager (32) + desc (64) +
+    // claim_holding_count (4) + padding (4) + StorageGap<2> (64). Total
+    // account size = 8 + 176 = 184. Note: StorageGap<N> is [[u8; 32]; N], so
+    // StorageGap<2> = 64 bytes. The on-chain processor enforces this layout
+    // via `const _: () = assert!(zero_copy::data_end::<ValidatorClientRewards>() == 184);`.
+    const DATA_LEN: usize = 184;
```
martinsander00: if the on-chain struct layout ever shifts, this silently produces broken bytes, the `data_end == 184` const-assert only catches it at program build time, not at fork load. can we build the account via the actual Pod type from the program crate? or at least a runtime size check

martinsander00 (approved):

PR #373, opened 2026-05-22
feat(solana-cli): tighten up some ergonomics (#1373)

bgm-malbeclabs:
## Summary of Changes
This PR incorporates feedback based on a prior review focused on tightening some ergonomics and sensible defaults.

Closes https://github.com/malbeclabs/infra/issues/1373

## Testing Verification
* A few tests have been updated to test the updated functionality


karl-dz (commented):
Tight PR. The shared `validate_shred_reward_token` helper is genuinely shared (configure batches the `ShredRewardToken` fetch into `get_multiple_accounts`, prepare-offchain-message reuses it via the async wrapper, no duplication). The `parse_valid_for` tests cover whitespace, zero, and overflow explicitly. The direct-path mismatch test asserts the `prepare-offchain-message` substring so the offchain workflow stays the lead remedy. The `show` command's switch to `try_fetch_zero_copy_data_with_commitment` plus `get_account_with_commitment(...).value` cleanly distinguishes RPC failure from absent account, and the `bytemuck::Zeroable` test fixture for `ShredRewardToken` avoids pulling `ruint` into the test crate.

My agents and I found 3 issues and 2 nits on this review.

crates/solana-client-tools/src/payer.rs
```diff
@@ -305,10 +305,13 @@ impl std::ops::Deref for Wallet {
     }
 }
 
-impl TryFrom<SolanaPayerOptions> for Wallet {
-    type Error = anyhow::Error;
-
-    fn try_from(opts: SolanaPayerOptions) -> Result<Wallet> {
+impl Wallet {
```
karl-dz: `try_new` should be moved into the existing `impl Wallet` block at line 83. As written this PR splits the type's method surface across two non-adjacent blocks, with `impl Deref for Wallet` between them. A reader looking at the original block has no signal that there are more methods 200 lines down, and the second block holds a single constructor that has no reason to live apart from `pubkey()`, `new_transaction_*`, etc. Folding it into the existing block keeps the API surface in one place.

crates/solana-cli/src/command/shreds/publisher_rewards/configure.rs
```diff
@@ -231,15 +234,21 @@ impl ConfigureCommand {
...
+        // Push the ATA-create only when the account doesn't already exist.
+        // The idempotent variant is the on-chain race-condition safety net
+        // (someone could race us between the read and the submit), not the
+        // primary mechanism — skipping it on the happy path keeps the
+        // transaction smaller and burns less compute. The fee-payer pays the
+        // rent; the account is owned by --rewards-token-owner.
+        if !ata_exists {
+            println!("Rewards ATA missing; will create as part of this transaction.");
+            instructions.push(create_associated_token_account_idempotent(
+                &wallet_key,
```
karl-dz: The funding address passed here is `wallet_key`, the signer's pubkey, but the comment above says "The fee-payer pays the rent". When `--fee-payer` is set the fee payer is a different keypair, and as written the ATA rent will still come from the signer's account. If an operator chose `--fee-payer` because the validator-identity keypair is rent-poor, this transaction will fail at submit.

This isn't really a problem this PR created. `Wallet` doesn't expose a helper for "give me the pubkey that pays this transaction's fees", so every call site has to re-derive it (and several existing ones likely have the same bug). A short-term fix here:

```rust
wallet.fee_payer.as_ref().map(Signer::pubkey).unwrap_or(wallet_key)
```

Longer term, a `Wallet::fee_payer_key(&self) -> Pubkey` helper would remove the footgun across the codebase. I will file a follow-up issue for that.

crates/solana-cli/src/command/shreds/publisher_rewards/configure.rs
```diff
@@ -231,15 +234,21 @@ impl ConfigureCommand {
...
+                &self.rewards_token_owner,
+                &rewards_token_mint,
+                &spl_token_interface::ID,
+            ));
+        }
 
         // CU budget: init (~20k) + configure (~20k baseline) + ed25519 verify
```
karl-dz: The CU budget here is a hardcoded 250k, but `find_shred_reward_token_address(&rewards_token_mint).0` and `find_validator_publisher_rewards_address(&self.node_id).0` discard the bumps. The established pattern across this crate (and a sibling at `crates/solana-cli/src/command/shreds/pay.rs:415,425`) builds the budget up from `Wallet::compute_units_for_bump_seed(bump)` per PDA derivation, with `const CREATE_ATA_CU_BASE: u32 = 25_000` already defined elsewhere in the workspace (`crates/contributor-rewards/src/calculator/distribute.rs:257` and `crates/solana-cli/src/command/revenue_distribution/relay/distribute_rewards.rs:217`).

The pattern was inherited from #360, so this is not strictly introduced by this PR. The conditional ATA-create added here makes the gap more visible: the budget is now a ~25k overestimate when the rewards ATA already exists, and the comment listing "create-ATA (~25k)" as part of the budget is now stale.

Could the bumps be captured (drop the `.0` discards) and the budget be rebuilt from `Wallet::compute_units_for_bump_seed(bump)` per PDA, with `CREATE_ATA_CU_BASE` added conditionally when `!ata_exists`?

I will file a follow-up issue proposing a helper that returns both the ATA pubkey and the CU cost for creating it, so future call sites get the right figure without re-deriving the bump.

crates/solana-cli/src/command/shreds/publisher_rewards/configure.rs
```diff
@@ -95,17 +91,19 @@ impl ResolvedAuth {
 
 /// Resolve `ResolvedAuth` from CLI inputs. Pure: no I/O, no network.
 ///
-/// - `node_id`: from `--node-id`, used to validate the direct-path fee-payer match.
-/// - `fee_payer_pubkey`: the pubkey of the wallet that pays for the transaction.
+/// - `node_id`: from `--node-id`, used to validate the direct-path signer match.
+/// - `signer_pubkey`: the pubkey of the keypair that signs the transaction
+///   (`-k`). With `--fee-payer` overriding the fee payer, this is the
```
karl-dz, line 97: Nit: with the rename to `signer_pubkey`, this docstring is the canonical place that explains the contract. A few comments elsewhere in the file still describe `-k` as the "fee-payer keypair":

- line 32, example header: "fee-payer keypair `-k` doubles as the validator identity"
- line 62, signature field doc: "the fee-payer keypair (`-k`) is used as the validator identity and signs the transaction directly"
- lines 260-261: "Single-signer transaction: the fee-payer keypair signs both as fee-payer and (in the direct path) as the validator identity"

These read a little oddly now that the parameter is `signer_pubkey`. Easy follow-up if you are touching the file.

crates/solana-cli/src/command/shreds/publisher_rewards/mod.rs
```diff
@@ -35,3 +41,133 @@ impl PublisherRewardsCommand {
...
+             pick an enabled mint or wait for the admin to re-enable it"
+        );
+    }
+    Ok(())
+}
+
+/// Fetch the `ShredRewardToken` PDA for `rewards_token_mint` and validate it
+/// (registered + enabled). Used by `prepare-offchain-message`; `configure`
+/// uses [`validate_shred_reward_token`] directly because it batches the fetch.
+pub(crate) async fn ensure_shred_reward_token_enabled(
```
karl-dz: Nit: `ensure_shred_reward_token_enabled` has one caller, and the doc-comment notes that `configure` calls `validate_shred_reward_token` directly because it batches the fetch. Could we inline the three-line `get_account_with_commitment` + `validate_shred_reward_token` sequence into `prepare_offchain_message.rs`? That would remove the asymmetry the doc-comment explains, and both call sites would use the same pattern.

karl-dz:
One more thing before merge: the PR description is still the template placeholders. Per the repo's GitHub conventions, the body should have a `Closes https://github.com/malbeclabs/infra/issues/1373` line and a summary covering the substantive changes.

bgm-malbeclabs:
> One more thing before merge: the PR description is still the template placeholders. Per the repo's GitHub conventions, the body should have a `Closes https://github.com/malbeclabs/infra/issues/1373` line and a summary covering the substantive changes.

It would be nice if Github would default to PRs as drafts. Don' t think I intended this to be reviewed yet

martinsander00:
> It would be nice if Github would default to PRs as drafts. Don' t think I intended this to be reviewed yet

if you create the pr with claude, you can tell it to pass the --draft flag when creating them

martinsander00 (dismissed):

bgm-malbeclabs:
> > It would be nice if Github would default to PRs as drafts. Don' t think I intended this to be reviewed yet
> 
> if you create the pr with claude, you can tell it to pass the --draft flag when creating them

i do it the old fashioned way - let me see if one can create a draft PR with the CLI that way

martinsander00 (approved):

PR #374, opened 2026-05-22
solana-client-tools: tolerate missing/invalid accounts in try_fetch_multiple_zero_copy_data

martinsander00:
## Summary

- Change `SolanaConnection::try_fetch_multiple_zero_copy_data` from `Result<Vec<T>>` to `Result<Vec<Option<T>>>`. Missing accounts AND accounts whose bytes fail discriminator/layout checks both surface as `None` in the corresponding slot, preserving the 1:1 mapping with input keys. A single bad slot no longer poisons the entire batch.
- Helper stays silent on `None` — callers with richer context (e.g. `device_key`, `history_key`) should log on their side.

## Motivation

Today, one missing or invalid PDA fails the whole batch via the strict `try_into()` path. Downstream consumers work around this by calling `try_fetch_multiple_accounts` directly and doing per-slot `checked_from_bytes_with_discriminator` themselves (see [doublezerofoundation/doublezero-shreds#418](https://github.com/malbeclabs/doublezero-shreds/pull/418) — a workaround in shred-oracle's `get_device_enabled_and_available_seats`, the read-side poll that was blocked on mainnet for hours by a single device activated without its `DeviceHistory` PDA).

Pushing the tolerance into the shared helper lets that caller (and any future one) go back to the one-liner. Companion PR in doublezero-shreds reverts the workaround and consumes this signature.

Closes [malbeclabs/infra#1366](https://github.com/malbeclabs/infra/issues/1366).

## In-tree callers

Only one: `validator-debt/src/rpc.rs` (Distribution batch). Migration is a single `.flatten()` between `.into_iter()` and the existing content-filter — which already tolerates un-finalized epochs, so the new "skip-missing" semantics matches its spirit.

## Testing

- `cargo check --workspace --all-targets` passes.
- `cargo test -p doublezero-solana-client-tools --lib` passes (13/13).


karl-dz (commented):
The change is tight: one helper signature, one in-tree caller, one CHANGELOG line. The new `and_then(|a| ZeroCopyAccountOwnedData::from_account(&a))` correctly handles the realistic "stray lamports at a future PDA" and "not yet initialized" cases without burning the whole batch, and the `Vec<Option<T>>` shape leaves callers free to pick fail-loud or flatten on a per-caller basis.

A few small notes on testing, the new docstring, and the changelog entry. One process note: this PR has the `skip-changelog` label but also adds a CHANGELOG entry. The label looks stale and could come off.

My agents and I found 1 issue and 2 nits on this review.

crates/solana-client-tools/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+- return vec of opt structs in try_fetch_multiple_zero_copy_data ([#374](https://github.com/doublezerofoundation/doublezero-offchain/pull/374))
```
karl-dz: Nit: this is a breaking public-API change (`Result<Vec<T>>` becomes `Result<Vec<Option<T>>>`), and the changelog line reads more like a commit subject than a user-facing note. Could we expand it to flag both the semantic change and the breakage? For example:

`tolerate missing and unparseable accounts in try_fetch_multiple_zero_copy_data. Return type is now Result<Vec<Option<_>>> (breaking)`

crates/solana-client-tools/src/rpc.rs
```diff
@@ -211,15 +211,20 @@ impl SolanaConnection {
         Ok(account_infos)
     }
 
+    /// Returns one slot per input key. Missing accounts and accounts whose
+    /// bytes fail discriminator/layout checks both surface as `None`, so a
+    /// single bad slot can't poison the entire batch. Callers that need to
+    /// distinguish missing-vs-invalid or that want to fail-loud should
+    /// handle the `None`s themselves (and log with their own context).
```
karl-dz: Nit on the docstring: two small refinements worth folding in.

- The line about callers needing to distinguish missing-vs-invalid implies they can. From the return value they cannot: both reasons collapse to `None`. Could you either drop that sentence or rephrase to acknowledge that the distinction is unrecoverable from the return value? A caller who needs it has to re-fetch the raw account.
- Perhaps we add a sentence about the caller's responsibility for account-level security. The helper checks discriminator and layout only. Anything else the caller cares about (account ownership, semantic validity of the parsed contents, whether the key was even expected to exist) is on the caller.

crates/solana-client-tools/src/rpc.rs
```diff
@@ -211,15 +211,20 @@ impl SolanaConnection {
...
         keys: &[Pubkey],
-    ) -> Result<Vec<ZeroCopyAccountOwnedData<T>>> {
-        self.try_fetch_multiple_accounts(keys)
+    ) -> Result<Vec<Option<ZeroCopyAccountOwnedData<T>>>> {
+        Ok(try_fetch_multiple_accounts(&self.0, keys)
             .await?
             .into_iter()
-            .map(TryInto::try_into)
-            .collect()
+            .map(|opt| opt.and_then(|a| ZeroCopyAccountOwnedData::from_account(&a)))
```
karl-dz, line 230: `ZeroCopyAccountOwnedData::from_account` has not had test coverage since it was introduced, and the changed helper now leans on its `None`-on-bad-bytes contract more pointedly than before. Could this PR be the natural moment to add the tests that should have been there from the start? A small `#[cfg(test)] mod tests` block in `zero_copy.rs` covering the four states `from_account` is expected to handle:

- Correct discriminator and correctly-sized data returns `Some`.
- Wrong discriminator returns `None`.
- Too-short data returns `None`.
- Empty data (the lamport-send case) returns `None`.

That pins down the contract `try_fetch_multiple_zero_copy_data` now relies on. A future change that tightens or relaxes `checked_from_bytes_with_discriminator` would otherwise break the helper silently.

crates/solana-client-tools/CHANGELOG.md
martinsander00: https://github.com/doublezerofoundation/doublezero-offchain/pull/374/changes/66e1904f3085c03560843d26f9ae6b371433eebc

crates/solana-client-tools/src/rpc.rs
martinsander00: https://github.com/doublezerofoundation/doublezero-offchain/pull/374/changes/66e1904f3085c03560843d26f9ae6b371433eebc

crates/solana-client-tools/src/rpc.rs
martinsander00: https://github.com/doublezerofoundation/doublezero-offchain/pull/374/changes/66e1904f3085c03560843d26f9ae6b371433eebc

bgm-malbeclabs (approved):

karl-dz (approved):
Merge commit made me twitch, but it's fine

martinsander00:
> Merge commit made me twitch, but it's fine

haha i resolved the conflict using github ui

PR #375, opened 2026-05-26
solana-cli: distribute pending rewards on validator configure

martinsander00:
Note: this PR carries some vendored snippets from `doublezero-shreds` that will be deleted once shreds merges into the monorepo — each vendored file carries a header tagging "remove when shreds merges into monorepo" so the duplication is greppable and bounded in time.

## Summary of changes

- After a successful `configure`, the CLI scans the last **100** subscription epochs and, for each `(epoch, mint)` pair where this validator's leaf bit is still set in the publisher accumulation bitmap, submits a `DistributeValidatorRewards` ix. Probes all three mint candidates (2Z, USDC, WSOL) per epoch to cover rewards routed historically against a different mint than the validator's current VPR config. Per-`(epoch, mint)` soft-fail; skipped under dry-run.

- Multi-client-id-aware: a validator that legitimately appears under multiple `client_id`s in one epoch (the leaf schema's dedup key is `(node_id, client_id)`, not `node_id` alone) emits one candidate per match, each with its own merkle proof and bitmap slot. The `InitializeClaimHolding` dedup is keyed by `(epoch, client_id)` so each per-client_id claim holding gets its own init.

- **Five batched upfront RPCs + one parallel fan-out** for the whole pass regardless of window size: `ShredDistribution`, journals across the three mints, parent `Distribution` (deduped by DZ epoch), claim holdings, candidate destination ATAs. S3 leaf fetches use `futures::buffer_unordered(8)` so a 100-epoch lookback costs ~1-3 s of S3 wall-time instead of 5-20 s sequential.

- **Per-tx hygiene**: each `(epoch, mint)` tx skips the redundant `CheckCliVersion` ix (configure already enforced it for the session), skips `create_associated_token_account_idempotent` when the destination ATA is known to exist (pre-checked + tracked across the pass), and reuses one cached blockhash instead of fetching per tx.

- The pass is fault-tolerant of its own RPC churn: if the distribute pass fails after `configure` has already landed, the CLI prints the error to stderr but returns `Ok` — operators don't see a misleading "configure failed" exit code over a downstream transient.

- Current Solana epoch is resolved via a Solana RPC (`getEpochInfo`), not the DZ-Ledger one that hosts the shred-subscription program — on testnet/localnet those are distinct chains with independent epoch numbers, and the pass needs the Solana epoch to match S3 file names and on-chain `subscription_epoch` PDAs.

- **Counter rename + new summary**: the pass exposes three counters with explicit units (`submits_succeeded` / `submits_failed` per `(epoch, mint)` submission; `epochs_unsettled` per-epoch where no submit landed for this leaf). The end-of-pass summary, printed at the configure call site:

  ```
  Distribute pass complete:
    submits: N succeeded, M failed
    epochs:  K still unsettled for this leaf
  ```

## SDK additions

- `DistributeValidatorRewards` ix variant + account builder. The builder's `client_mint_key` is always-required `&Pubkey`; the omit-rule (when client journal == publisher journal, drop the redundant client journal account) is decided once inside the builder instead of overloaded onto an `Option`. Borsh `Serialize` + `Deserialize` round-trip with a frozen-bytes test.
- Zero-copy mirrors for `ShredDistribution`, `ShredDistributionJournal`, `ValidatorClientRewardsConfig`, each with a compile-time `const _: () = assert!(size_of::<T>() == ...)` matching the on-chain `data_end::<T>()` assert. Any silent upstream layout drift breaks the offchain build instead of producing garbage bitmap reads at runtime.
- Journal PDA finder, `ValidatorRewardsLeaf` Pod struct.

## Vendored snippets

- S3 fetch + merkle helper from `doublezero-shreds` (~190 LOC). Parity-critical: the canonical oracle uses the same silent `filter_map(...).ok()?` for unparseable `node_identity` entries, and the merkle root must match exactly or every proof fails on-chain verification. Pinned with a `compute_leaves_drops_unparseable_pubkeys_silently` test + load-bearing-parity doc block. `tracing::warn!` added (offchain-only) so operators see when an entry is silently dropped.
- `ValidatorRewardsLeaf` Pod struct, byte-for-byte compatible with on-chain.
- The vendored `compute_leaves_with_proofs` (O(N log N) all-proofs builder) is intentionally NOT carried over — offchain uses a `compute_proof_for_leaf` helper that runs O(log N) work per matched leaf. A `compute_proof_for_leaf_reconstructs_root` test pins equivalence with the on-chain root.

## Bitmap convention parity

The accumulation-bitmap read (`bitmap[leaf_index / 8] & (1 << (leaf_index % 8))`) mirrors what on-chain `try_process_remaining_data_leaf_index` writes (LSB-first via `ByteFlags::set_bit`). Pinned with a `bitmap_bit_set_matches_onchain_accumulate_convention` test that builds a bitmap by calling an on-chain-style write helper and asserts the offchain reader returns the expected value across byte boundaries.

## Testing verification

- 138 CLI unit tests + 35 SDK unit tests pass.
- New tests added for: bitmap-convention parity, the `compute_leaves` silent-drop parity pin, single-leaf merkle proof equivalence, the `DistributeValidatorRewards` ix Borsh round-trip, and the layout-size asserts (which run at compile time).
- `cargo fmt --check` clean, `cargo clippy --all-targets` clean across workspace.
- No devnet smoke yet — last step before merge is running `publisher-rewards configure` against a real validator with pending epochs on devnet to confirm the end-to-end flow lands.

### DIFF
--- a/crates/solana-cli/CHANGELOG.md
+++ b/crates/solana-cli/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- `shreds publisher-rewards configure`: after the configure tx lands, scan the last 20 subscription epochs and submit `DistributeValidatorRewards` for any unsettled leaf so the validator's pending rewards land in the same operator session. Skipped under dry-run; per-epoch soft-fail so one bad epoch doesn't tank the rest (#375)
+
 ## [0.5.5](https://github.com/doublezerofoundation/doublezero-offchain/releases/tag/doublezero-solana/v0.5.5)
 
 - uptick crate to v0.5.5 (#372)
--- a/crates/solana-cli/CHANGELOG.md
+++ b/crates/solana-cli/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- `shreds publisher-rewards configure`: after the configure tx lands, scan the last 20 subscription epochs and submit `DistributeValidatorRewards` for any unsettled leaf so the validator's pending rewards land in the same operator session. Skipped under dry-run; per-epoch soft-fail so one bad epoch doesn't tank the rest (#375)
+- `shreds publisher-rewards configure`: after the configure tx lands, scan the last 100 subscription epochs and submit `DistributeValidatorRewards` for any unsettled leaf so the validator's pending rewards land in the same operator session. Skipped under dry-run; per-epoch soft-fail so one bad epoch doesn't tank the rest (#375)
 
 ## [0.5.5](https://github.com/doublezerofoundation/doublezero-offchain/releases/tag/doublezero-solana/v0.5.5)
 
--- a/crates/solana-cli/src/command/shreds/publisher_rewards/configure.rs
+++ b/crates/solana-cli/src/command/shreds/publisher_rewards/configure.rs
@@ -336,10 +336,8 @@ impl ConfigureCommand {
             match distribute_result {
                 Ok(outcome) => {
                     println!(
-                        "\nDistribute pass complete:\n  \
-                         submits: {} succeeded, {} failed\n  \
-                         epochs:  {} still unsettled for this leaf",
-                        outcome.submits_succeeded, outcome.submits_failed, outcome.epochs_unsettled,
+                        "\nDistribute pass complete: {} distributed, {} failed.",
+                        outcome.distributed, outcome.failed,
                     );
                 }
                 Err(error) => {
--- a/crates/solana-cli/src/command/shreds/publisher_rewards/configure.rs
+++ b/crates/solana-cli/src/command/shreds/publisher_rewards/configure.rs
@@ -305,9 +305,31 @@ impl ConfigureCommand {
         let transaction = wallet.new_transaction(&instructions).await?;
 
         let tx_outcome = wallet.send_or_simulate_transaction(&transaction).await?;
+        let configure_executed = matches!(tx_outcome, TransactionOutcome::Executed(_));
         if let TransactionOutcome::Executed(tx_sig) = tx_outcome {
             println!("Configured validator publisher rewards: {tx_sig}");
             wallet.print_verbose_output(&[tx_sig]).await?;
+        }
+
+        // Post-configure: distribute this validator's pending rewards for
+        // any recent subscription epoch where accumulation has completed
+        // but distribute hasn't run for this leaf yet. Skipped under
+        // dry-run (the configure tx never actually wrote the VPR state we
+        // would distribute under).
+        if configure_executed {
+            let network_env = wallet
+                .connection
+                .try_network_environment()
+                .await
+                .context("detecting network environment")?;
+            super::distribute::try_distribute_pending(
+                &wallet,
+                &self.node_id,
+                &self.rewards_token_owner,
+                &rewards_token_mint,
+                network_env,
+            )
+            .await?;
         }
 
         Ok(())
--- a/crates/solana-cli/src/command/shreds/publisher_rewards/configure.rs
+++ b/crates/solana-cli/src/command/shreds/publisher_rewards/configure.rs
@@ -314,22 +314,40 @@ impl ConfigureCommand {
         // Post-configure: distribute this validator's pending rewards for
         // any recent subscription epoch where accumulation has completed
         // but distribute hasn't run for this leaf yet. Skipped under
-        // dry-run (the configure tx never actually wrote the VPR state we
-        // would distribute under).
+        // dry-run (the configure tx never actually wrote the ValidatorPublisherRewards
+        // state we would distribute under).
         if configure_executed {
-            let network_env = wallet
-                .connection
-                .try_network_environment()
+            let distribute_result = async {
+                let network_env = wallet
+                    .connection
+                    .try_network_environment()
+                    .await
+                    .context("detecting network environment")?;
+                super::distribute::try_distribute_pending(
+                    &wallet,
+                    &solana_connection,
+                    &self.node_id,
+                    &self.rewards_token_owner,
+                    network_env,
+                )
                 .await
-                .context("detecting network environment")?;
-            super::distribute::try_distribute_pending(
-                &wallet,
-                &self.node_id,
-                &self.rewards_token_owner,
-                &rewards_token_mint,
-                network_env,
-            )
-            .await?;
+            }
+            .await;
+            match distribute_result {
+                Ok(outcome) => {
+                    println!(
+                        "\nDistribute pass complete:\n  \
+                         submits: {} succeeded, {} failed\n  \
+                         epochs:  {} still unsettled for this leaf",
+                        outcome.submits_succeeded, outcome.submits_failed, outcome.epochs_unsettled,
+                    );
+                }
+                Err(error) => {
+                    eprintln!(
+                        "\nDistribute pass failed (configure already landed; safe to re-run): {error:#}"
+                    );
+                }
+            }
         }
 
         Ok(())
--- a/crates/solana-cli/src/command/shreds/publisher_rewards/distribute.rs
+++ b/crates/solana-cli/src/command/shreds/publisher_rewards/distribute.rs
@@ -22,7 +22,6 @@ use doublezero_solana_client_tools::{
     account::zero_copy::ZeroCopyAccountOwnedData,
     payer::{TransactionOutcome, Wallet},
     rpc::{NetworkEnvironment, SolanaConnection},
-    transaction::try_new_transaction,
 };
 use doublezero_solana_sdk::{
     Pubkey, environment_2z_token_mint_key, environment_usdc_token_mint_key,
@@ -47,10 +46,7 @@ use doublezero_solana_sdk::{
     try_build_instruction,
 };
 use futures::stream::{self, StreamExt};
-use solana_sdk::{
-    compute_budget::ComputeBudgetInstruction, hash::Hash, instruction::Instruction,
-    signature::Keypair, signer::Signer,
-};
+use solana_sdk::{compute_budget::ComputeBudgetInstruction, instruction::Instruction};
 use spl_associated_token_account_interface::{
     address::get_associated_token_address, instruction::create_associated_token_account_idempotent,
 };
@@ -325,7 +321,7 @@ pub async fn try_distribute_pending(
         .await
         .context("fetching claim holding accounts")?;
 
-    // ----- Step 6: pre-fetch destination ATAs + cache the blockhash -----
+    // ----- Step 6: pre-fetch destination ATAs -----
     //
     // Across the entire pass there are at most three distinct
     // destination ATAs (one per candidate reward mint:
@@ -335,12 +331,6 @@ pub async fn try_distribute_pending(
     // the destination is actually missing — saves ~25k CU per tx that
     // would otherwise pay the SPL Token existence check on a no-op
     // create. The set is mutated as freshly-created ATAs land.
-    //
-    // The blockhash is also cached once. `new_transaction` normally
-    // calls `getLatestBlockhash` per tx; for ~300 distribute txs that's
-    // 300 redundant RPCs. Blockhashes stay valid for ~60-90 s, well
-    // above this pass's wall time; on `BlockhashNotFound` at submit the
-    // operator re-runs (the pass is idempotent).
     let candidate_destination_atas: Vec<Pubkey> = journal_mint_candidates
         .iter()
         .map(|mint| get_associated_token_address(rewards_token_owner_key, mint))
@@ -355,12 +345,6 @@ pub async fn try_distribute_pending(
         .zip(candidate_destination_ata_accounts.iter())
         .filter_map(|(ata, account)| (!account.data.is_empty()).then_some(*ata))
         .collect();
-
-    let cached_blockhash = wallet
-        .connection
-        .get_latest_blockhash()
-        .await
-        .context("fetching cached blockhash for distribute pass")?;
 
     // ----- Step 7: in-memory decision loop, submit per (epoch, mint) -----
     //
@@ -452,7 +436,6 @@ pub async fn try_distribute_pending(
                 &dz_mint_key,
                 needs_init,
                 needs_ata_create,
-                cached_blockhash,
             )
             .await
             {
@@ -499,7 +482,6 @@ async fn submit_distribute_tx(
     dz_mint_key: &Pubkey,
     needs_init: bool,
     needs_ata_create: bool,
-    recent_blockhash: Hash,
 ) -> Result<()> {
     let mut instructions: Vec<Instruction> = Vec::new();
 
@@ -565,24 +547,11 @@ async fn submit_distribute_tx(
         instructions.push(compute_unit_price_ix.clone());
     }
 
-    // Inlined tx build (replaces `wallet.new_transaction`) so we can reuse
-    // the cached blockhash and skip the per-tx `getLatestBlockhash` RPC.
-    // Signer assembly mirrors `Wallet::new_transaction_with_additional_signers_and_lookup_tables`:
-    // when `--fee-payer` is set it signs first; the `-k` signer is added
-    // only if its pubkey differs from the fee payer's.
-    let mut signers: Vec<&Keypair> = Vec::with_capacity(2);
-    match wallet.fee_payer {
-        Some(ref fee_payer) => {
-            signers.push(fee_payer);
-            if wallet.signer.pubkey() != fee_payer.pubkey() {
-                signers.push(&wallet.signer);
-            }
-        }
-        None => {
-            signers.push(&wallet.signer);
-        }
-    }
-    let transaction = try_new_transaction(&instructions, &signers, &[], recent_blockhash)?;
+    // Fetch a fresh blockhash per tx. The pass submits sequentially and a
+    // validator with many pending epochs can run long enough that a single
+    // blockhash cached up front ages out of the cluster's validity window
+    // mid-pass — every later tx then fails with "Blockhash not found".
+    let transaction = wallet.new_transaction(&instructions).await?;
     let tx_outcome = wallet.send_or_simulate_transaction(&transaction).await?;
     if let TransactionOutcome::Executed(tx_sig) = tx_outcome {
         println!(
--- a/crates/solana-cli/src/command/shreds/publisher_rewards/distribute.rs
+++ b/crates/solana-cli/src/command/shreds/publisher_rewards/distribute.rs
@@ -38,8 +38,9 @@ use doublezero_solana_sdk::{
         },
         state::{
             ShredDistribution, ShredDistributionJournal, find_claim_holding_address,
-            find_shred_distribution_address, find_shred_distribution_journal_address,
-            find_validator_client_rewards_address,
+            find_program_config_address, find_shred_distribution_address,
+            find_shred_distribution_journal_address, find_validator_client_rewards_address,
+            is_distribute_validator_rewards_enabled,
         },
         types::ValidatorRewardsLeaf,
     },
@@ -99,6 +100,24 @@ pub async fn try_distribute_pending(
     rewards_token_owner_key: &Pubkey,
     network_env: NetworkEnvironment,
 ) -> Result<DistributeOutcome> {
+    let mut outcome = DistributeOutcome::default();
+
+    // Check if distribute flag is enabled in ProgramConfig.
+    let program_config_account = wallet
+        .connection
+        .try_fetch_multiple_accounts(&[find_program_config_address().0])
+        .await
+        .context("fetching program config")?;
+    let distribute_enabled = program_config_account
+        .first()
+        .is_some_and(|account| is_distribute_validator_rewards_enabled(&account.data));
+    if !distribute_enabled {
+        println!(
+            "\nDistribute is not enabled on this cluster yet; skipping pending-rewards distribution."
+        );
+        return Ok(outcome);
+    }
+
     // `wallet.connection` is the DZ-Ledger RPC (program host). Its epoch
     // is the DZ-Ledger epoch, which has no relation to the Solana epoch
     // on testnet/localnet. `subscription_epoch` PDAs and S3 file names
@@ -121,8 +140,6 @@ pub async fn try_distribute_pending(
     let usdc_mint_key = environment_usdc_token_mint_key(network_env);
     let wsol_mint_key = spl_token_interface::native_mint::ID;
     let journal_mint_candidates = [dz_mint_key, usdc_mint_key, wsol_mint_key];
-
-    let mut outcome = DistributeOutcome::default();
 
     // ----- Step 1: batch fetch ShredDistribution accounts in the window -----
 
@@ -454,8 +471,10 @@ pub async fn try_distribute_pending(
                     }
                 }
                 Err(error) => {
+                    let full = format!("{error:#}");
+                    let summary = full.lines().next().unwrap_or(&full);
                     eprintln!(
-                        "  epoch {} mint {publisher_mint}: failed: {error:#}",
+                        "  epoch {} mint {publisher_mint}: failed: {summary}",
                         candidate.subscription_epoch
                     );
                     outcome.submits_failed += 1;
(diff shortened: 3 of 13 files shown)


crates/solana-cli/src/command/shreds/publisher_rewards/distribute.rs
```diff
@@ -0,0 +1,494 @@
...
+    wallet: &Wallet,
+    node_id: &Pubkey,
+    rewards_token_owner_key: &Pubkey,
+    rewards_token_mint_key: &Pubkey,
+    network_env: NetworkEnvironment,
+) -> Result<DistributeOutcome> {
+    let current_epoch = wallet
+        .connection
+        .0
+        .get_epoch_info()
```
bgm-malbeclabs, line 124: **CONFIRMED, HIGH — wrong epoch source on testnet/localnet.** `wallet.connection` is built via `into_shred_subscription_connection()` (configure.rs:165), which on testnet resolves to `PUBLIC_DOUBLEZERO_LEDGER_TESTNET_URL` (rpc.rs:65). `wallet.connection.0.get_epoch_info().epoch` therefore returns the **DZ-Ledger epoch**, not the Solana epoch — yet `subscription_epoch` PDAs and S3 file names (`{solana_epoch}.json`) are Solana-epoch keyed.

On testnet the lookback derives `find_shred_distribution_address` PDAs for epoch numbers that don't exist, `accumulated` is empty, and the pass prints `No accumulated epochs in the window` — all pending rewards silently skipped. Mainnet works only because `shred_subscription_url(MainnetBeta) == solana_public_url(MainnetBeta)`.

Fix: query the epoch from a Solana RPC. `configure.rs:157` already constructs a `SolanaConnection` from the same options; thread that in (or take it as a separate parameter).

crates/solana-sdk/src/shred_subscription/state.rs
```diff
@@ -683,6 +699,181 @@ impl PrecomputedDiscriminator for ValidatorPublisherRewards {
...
+#[repr(C, align(8))]
+pub struct ValidatorClientRewardsConfig {
+    pub default_proportion: UnitShare16,
+    _padding_0: [u8; 2],
+    pub proportions: ValidatorClientRewardProportions,
+}
+
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq, Pod, Zeroable)]
+#[repr(C, align(8))]
+pub struct ShredDistribution {
```
bgm-malbeclabs, line 770: **CONFIRMED, HIGH — no layout parity assertion for the new vendored Pod structs.** This same file already pairs `ValidatorClientRewards` with `VCR_ACCOUNT_DATA_LEN = 184` (line 605) and the comment _"Mirrors the program's `assert!(zero_copy::data_end::<ValidatorClientRewards>() == 184)`. Update both sides together if the on-chain layout changes."_ The new `ShredDistribution` / `ShredDistributionJournal` / `ValidatorClientRewardsConfig` blocks have no equivalent pin.

The risk class is real: the project already documents one runtime failure where `StorageGap` was sized wrong. Upstream drift in any field (or in `Flags`/`StorageGap<N>`'s size against a different `program-tools` pin) silently shifts `remaining_data`'s start — and `ShredDistributionJournal.publisher_accumulation_bitmap_{start,end}_index` indexes into `remaining_data`. Result: bitmap reads land on wrong bytes, distribute silently undercounts or duplicates work.

Minimum fix: add `const _: () = assert!(std::mem::size_of::<ShredDistribution>() == EXPECTED && std::mem::size_of::<ShredDistributionJournal>() == EXPECTED && std::mem::size_of::<ValidatorClientRewardsConfig>() == EXPECTED);` populated from the upstream repo's own asserts. Better: pair with at least one bytes-fixture round-trip test.

crates/solana-cli/src/command/shreds/publisher_rewards/configure.rs
```diff
@@ -305,11 +305,33 @@ impl ConfigureCommand {
...
+        // but distribute hasn't run for this leaf yet. Skipped under
+        // dry-run (the configure tx never actually wrote the VPR state we
+        // would distribute under).
+        if configure_executed {
+            let network_env = wallet
+                .connection
+                .try_network_environment()
+                .await
+                .context("detecting network environment")?;
+            super::distribute::try_distribute_pending(
```
bgm-malbeclabs: **CONFIRMED, HIGH — hard-fail of `configure` after the configure tx already landed.** The configure tx signature was printed on line 311, but `try_distribute_pending(...).await?` propagates ANY error from network-env detection (line 322), `get_epoch_info`, the four `try_fetch_multiple_accounts` calls, S3 client construction, etc. as the CLI's exit code.

Failure mode: CI runs `configure`, the tx lands, then a transient RPC blip during the distribute pass turns the whole command into a non-zero exit. The operator/CI sees `configure` as failed and retries — `configure` is idempotent on-chain so it survives, but the operator is misled about what actually happened.

Fix: catch and log the result instead of `?`. The configure tx already landed; downstream failures shouldn't change the command's exit status.

crates/solana-cli/src/command/shreds/publisher_rewards/distribute.rs
```diff
@@ -0,0 +1,494 @@
...
+                        "  epoch {} mint {publisher_mint}: failed: {error:#}",
+                        candidate.subscription_epoch
+                    );
+                    outcome.failed += 1;
+                }
+            }
+        }
+
+        if !any_distributed_this_epoch {
+            outcome.skipped += 1;
```
bgm-malbeclabs: **CONFIRMED, MED — `outcome.skipped` double-counts on all-mints-failed epochs.** When all 3 per-mint distribute submits fail for one epoch, `outcome.failed += 1` runs 3 times AND this `outcome.skipped += 1` also fires (because `any_distributed_this_epoch` stayed false). Summary prints `failed=3, skipped=1` for what is really a single epoch with three failed submits.

Fix: gate `skipped` on the absence of failures too (e.g., track `any_failed_this_epoch` alongside `any_distributed_this_epoch`), or split the counter semantics (`epochs_skipped` vs `submits_failed`).

crates/solana-cli/src/command/shreds/publisher_rewards/distribute.rs
```diff
@@ -0,0 +1,494 @@
...
+            Err(err) => {
+                eprintln!("  epoch {epoch}: failed to compute merkle proofs: {err:#}");
+                outcome.failed += 1;
+                continue;
+            }
+        };
+        let Some(leaf_index) = leaves
+            .leaves
+            .iter()
+            .position(|leaf| &leaf.node_id == node_id)
```
bgm-malbeclabs: **PLAUSIBLE, MED — multi-client-id validator strands rewards.** `.position(...)` returns only the first match by `node_id`. The leaf schema is `(node_id, leader_slots, client_id)` and `compute_leaves` only rejects duplicate `(node_id, client_id)` pairs — multiple leaves with the same `node_id` but different `client_id` are explicitly legal (the sort key includes `client_id` for exactly that reason).

If a validator legitimately participates under multiple client_ids in one epoch (50 slots for client_id=1, 50 for client_id=2), this pass distributes for client_id=1 only — the other half is never touched by this validator's post-configure run.

Fix: iterate all leaves where `leaf.node_id == *node_id` and emit one `Candidate` per match.

crates/solana-cli/src/command/shreds/publisher_rewards/configure.rs
```diff
@@ -305,11 +305,33 @@ impl ConfigureCommand {
...
             println!("Configured validator publisher rewards: {tx_sig}");
             wallet.print_verbose_output(&[tx_sig]).await?;
         }
 
+        // Post-configure: distribute this validator's pending rewards for
+        // any recent subscription epoch where accumulation has completed
+        // but distribute hasn't run for this leaf yet. Skipped under
+        // dry-run (the configure tx never actually wrote the VPR state we
+        // would distribute under).
+        if configure_executed {
```
bgm-malbeclabs, line 319: **PLAUSIBLE, MED — altitude / code-judo.** Distribute is a hidden side-effect of `configure`. `publisher_rewards/mod.rs` already declares `pub mod distribute;` but no `Distribute` subcommand variant exists. Promote it to a real subcommand (e.g. `publisher-rewards distribute --node-id X --rewards-token-owner Y`) and have `configure` either call it through the same path or just print `now run distribute`.

Upside: (1) operators get a retry path after a transient distribute failure without paying for another configure tx; (2) the feature becomes testable in isolation; (3) this `if configure_executed { ... }` block and the redundant double-match on `tx_outcome` at lines 308/309 collapse — one match arm, no shim.

The `let configure_executed = matches!(tx_outcome, Executed(_));` immediately followed by `if let Executed(tx_sig) = tx_outcome` matches the same variant twice — a smell that the surrounding logic wants to be one thing, not two.

bgm-malbeclabs (commented):
Thorough review focused on correctness, structural simplification, and vendoring parity. Verdict: feature is functional and well-commented, but ships with two confirmed correctness bugs that affect testnet/localnet operation and the operator's exit-status reading of `configure`, plus a vendoring-parity hole that mirrors a previously documented incident class. There's also a clear code-judo move (promote `distribute` to a real subcommand) that would collapse much of the awkwardness around the `if configure_executed` shim. Inline comments below, ranked roughly by severity.

**Highest-leverage fixes**: (1) derive `current_epoch` from a Solana RPC, not the DZ-Ledger one; (2) follow the existing `VCR_ACCOUNT_DATA_LEN == 184` precedent for the new vendored Pod structs; (3) don't `?` propagate distribute errors after the configure tx already landed; (4) promote `distribute` to a real subcommand.

crates/solana-cli/src/command/shreds/publisher_rewards/distribute.rs
```diff
@@ -0,0 +1,494 @@
...
+            "  epoch {} mint {publisher_mint_key}: distributed ({tx_sig})",
+            candidate.subscription_epoch
+        );
+        wallet.print_verbose_output(&[tx_sig]).await?;
+    }
+
+    Ok(())
+}
+
+fn bitmap_bit_set(bitmap: &[u8], leaf_index: usize) -> bool {
```
bgm-malbeclabs, line 567: **PLAUSIBLE, MED — bitmap convention vs on-chain accumulate is unverified.** This helper assumes (a) LSB-first within each byte (bit 0 in byte 0 is leaf 0), and (b) that `leaf_index` maps directly to bit position 0 of the slice — i.e., that `publisher_accumulation_bitmap_start_index` is the absolute byte offset and the bitmap inside the slice starts at logical leaf 0. There's no parity test or shared helper with the on-chain accumulate ix.

Failure mode is silent: if on-chain uses MSB-first or the bitmap is logically offset, every check returns the wrong bit; `bitmap_bit_set` returns false; the candidate is skipped; no diagnostic.

Fix: construct a journal with known bits set and assert `bitmap_bit_set` matches what the on-chain accumulate logic would produce. Ideally reuse a shared helper from the on-chain code rather than reimplementing the convention locally.

Aside: `bitvec` is already a workspace dependency (`contributor-rewards`); consider `bitmap.view_bits::<Lsb0>().get(leaf_index)` over the open-coded `>>` if reuse is preferred.

crates/solana-cli/src/command/shreds/publisher_rewards/s3.rs
```diff
@@ -0,0 +1,210 @@
...
+    debug!(count = entries.len(), "Fetched validator entries");
+    Ok(entries)
+}
+
+pub fn compute_leaves(entries: &[ValidatorLeaderSlotEntry]) -> Result<ComputedLeaves> {
+    ensure!(!entries.is_empty(), "no validator entries to compute root");
+
+    let mut leaves = entries
+        .iter()
+        .filter_map(|entry| {
```
bgm-malbeclabs: **PLAUSIBLE, MED — `compute_leaves` silently drops unparseable Pubkeys.** `filter_map` skips any entry whose `node_identity` doesn't parse. On-chain accumulate works from the same JSON; if it doesn't apply the exact same filter (or applies a different one), the merkle leaf sets diverge, every reconstructed proof yields a different root, and ALL `DistributeValidatorRewards` txs for that epoch fail on-chain proof verification.

The failure looks like a chain bug rather than a parser disagreement — the operator sees opaque proof rejections, not a clear divergence signal.

Fix: replace with `ensure!` (hard-fail on parse error, mirroring whatever the canonical oracle impl does upstream), or document the filter as load-bearing parity and pin it in a test.

crates/solana-cli/src/command/shreds/publisher_rewards/s3.rs
```diff
@@ -0,0 +1,210 @@
...
+    entries: &[ValidatorLeaderSlotEntry],
+) -> Result<DistributionLeaves> {
+    let ComputedLeaves {
+        leaves,
+        root,
+        total_publishing_validators,
+        total_published_leader_slots,
+    } = compute_leaves(entries)?;
+
+    let proofs = (0..total_publishing_validators)
```
bgm-malbeclabs: **CONFIRMED, LOW-MED — wasted O(N log N) merkle proof work.** This loop builds a `MerkleProof` for every leaf in the tree, but the only caller in `distribute.rs:151` uses just `proofs[leaf_index]`. For N validators per epoch × 100 epochs in the lookback, that's tens of millions of avoidable SHA-256 ops per configure call.

Fix: expose a `compute_proof_for(entries, leaf_index)` (or `ComputedLeaves::proof_for(i)`) variant that runs one proof. Keep `compute_leaves_with_proofs` if it's needed elsewhere, but don't make the offchain CLI pay for proofs it discards.

crates/solana-cli/src/command/shreds/publisher_rewards/distribute.rs
```diff
@@ -0,0 +1,494 @@
...
+        println!("No accumulated epochs in the window; nothing to distribute.");
+        return Ok(outcome);
+    }
+
+    // ----- Step 2: fan out S3 to find this validator's leaf per epoch -----
+
+    let s3_client = s3::build_s3_client()?;
+    let mut candidates: Vec<Candidate> = Vec::new();
+    for (epoch, _shred_distribution) in &accumulated {
+        let entries = match s3::fetch_leader_slot_data(&s3_client, *epoch).await {
```
bgm-malbeclabs: **CONFIRMED, LOW-MED — sequential S3 fetches across the entire lookback window.** Strict serial WAN round-trips: 100 epochs × ~50–200ms each = 5–20s of avoidable wall time per configure.

Fix: `futures::stream::iter(&accumulated).map(|(e, _)| s3::fetch_leader_slot_data(&s3_client, *e)).buffer_unordered(8).collect().await` would cut this ~8x while still being polite to S3.

crates/solana-cli/src/command/shreds/publisher_rewards/distribute.rs
```diff
@@ -0,0 +1,494 @@
...
+
+    println!(
+        "\nDistribute pass complete: {} succeeded, {} skipped, {} failed.",
+        outcome.successful, outcome.skipped, outcome.failed,
+    );
+    Ok(outcome)
+}
+
+#[allow(clippy::too_many_arguments)]
+async fn submit_distribute_tx(
```
bgm-malbeclabs, line 475: **CONFIRMED, LOW-MED — per-tx redundant work.** Three sources of avoidable overhead per `(epoch, mint)` tx:

1. `build_check_cli_version_instruction()` prepended every tx (~5k CU each). The just-landed configure tx already enforced version compatibility — one check at the top of `try_distribute_pending` suffices.
2. `create_associated_token_account_idempotent` prepended every tx (~25k CU even when the ATA exists — the existence probe still costs CU). There are at most 3 distinct destination ATAs across the entire pass (one per journal mint); pre-check once via `getMultipleAccounts` before the loop.
3. `wallet.new_transaction(&instructions)` calls `get_latest_blockhash` per tx (payer.rs:138). With 300 txs that's 300 redundant `getLatestBlockhash` RPCs. Blockhashes are valid ~60-90s; cache and refresh only on `BlockhashNotFound`.

Combined: ~9M CU + 300 RPCs of avoidable work per pass.

Also: `DistributeValidatorRewardsAccounts::new` re-derives `find_program_config_address` and `find_validator_publisher_rewards_address(node_id)` per call — both constant across the pass. Hoist them.

crates/solana-cli/CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
```
bgm-malbeclabs, line 9: **CONFIRMED, LOW — CHANGELOG / code drift.** This line says _"scan the last **20** subscription epochs"_, but `distribute.rs:128` declares `DISTRIBUTE_LOOKBACK_EPOCHS: u64 = 100` (and its comment justifies 100 as the `getMultipleAccounts` 100-key chunk fit). Pure docs bug — reconcile.

crates/solana-cli/src/command/shreds/publisher_rewards/distribute.rs
```diff
@@ -0,0 +1,494 @@
...
+
+    // Mints we probe each epoch. We don't assume the validator's current
+    // VPR mint matches what they were configured against historically;
+    // accumulate may have routed earlier-epoch rewards into a different
+    // journal. Probing all three covers that case.
+    let dz_mint_key = environment_2z_token_mint_key(network_env);
+    let usdc_mint_key = environment_usdc_token_mint_key(network_env);
+    let wsol_mint_key = spl_token_interface::native_mint::ID;
+    let journal_mint_candidates = [dz_mint_key, usdc_mint_key, wsol_mint_key];
+    let _ = rewards_token_mint_key; // kept in signature for callers that want
```
bgm-malbeclabs: **CONFIRMED, LOW — discarded parameter + unused return type.** This `let _ = rewards_token_mint_key;` discards a parameter with a comment hedging _"kept in signature for callers that want to log it"_ — but the only caller (`configure.rs:329`) passes it in and never logs anything. The `DistributeOutcome` return is also dropped via `?` at the only call site.

Fix: drop the parameter from the signature; either log the outcome at the call site or change the return type to `()`. The signature should match what the function actually needs.

crates/solana-sdk/src/shred_subscription/instruction/account.rs
```diff
@@ -601,6 +601,161 @@ impl From<ConfigureValidatorPublisherRewardsAccounts> for Vec<AccountMeta> {
...
+    pub subscription_epoch: u64,
+    pub associated_dz_epoch: u64,
+    pub node_id: &'a Pubkey,
+    pub client_id: u16,
+    pub rewards_token_owner_key: &'a Pubkey,
+    pub publisher_mint_key: &'a Pubkey,
+    pub publisher_reward_mint_key: &'a Pubkey,
+    // 2Z mint for the standard configuration; `None` when the publisher
+    // journal also plays the client role (omit-rule).
+    pub client_mint_key: Option<&'a Pubkey>,
```
bgm-malbeclabs: **LOW (API design) — `Option<&Pubkey>` overloads "omit-rule fires".** Every caller has to encode the omit predicate itself (`if publisher_mint == dz_mint { None } else { Some(dz_mint) }` — see `distribute.rs:412`). The filter+`map`/filter+`unwrap_or` pattern then repeats in two slightly different shapes inside `DistributeValidatorRewardsAccounts::new` (lines ~653 and ~659).

Fix: take `client_mint_key: &Pubkey` always and have the builder make the omit decision once internally. One filter, one source of truth — callers don't need to know the rule.

crates/solana-cli/src/command/shreds/publisher_rewards/distribute.rs
```diff
@@ -0,0 +1,494 @@
...
+        let Some(leaf_index) = leaves
+            .leaves
+            .iter()
+            .position(|leaf| &leaf.node_id == node_id)
+        else {
+            // Validator wasn't a leader this epoch; silent skip (no work
+            // to do, not an error).
+            continue;
+        };
+        let shred_distribution = accumulated
```
bgm-malbeclabs: **CONFIRMED, LOW — pointless re-lookup.** The outer loop is `for (epoch, _shred_distribution) in &accumulated`, and this block then does `accumulated.iter().find(|(e, _)| e == epoch).map(|(_, sd)| sd).expect(...)` to fetch the same element by re-scanning the vector. The `_shred_distribution` binding IS the value being searched.

O(N²) over `accumulated` (up to 100 entries) plus a panic site that's structurally unreachable. Reads like leftover from a refactor — rename `_shred_distribution` → `shred_distribution` and use it directly.

crates/solana-cli/src/command/shreds/publisher_rewards/distribute.rs
martinsander00: https://github.com/doublezerofoundation/doublezero-offchain/pull/375/changes/321e23a441c4a74c6d60197a1a18dbf6edfd3ada

crates/solana-sdk/src/shred_subscription/state.rs
martinsander00: https://github.com/doublezerofoundation/doublezero-offchain/pull/375/changes/321e23a441c4a74c6d60197a1a18dbf6edfd3ada

crates/solana-cli/src/command/shreds/publisher_rewards/configure.rs
martinsander00: https://github.com/doublezerofoundation/doublezero-offchain/pull/375/changes/321e23a441c4a74c6d60197a1a18dbf6edfd3ada

crates/solana-cli/src/command/shreds/publisher_rewards/distribute.rs
martinsander00: https://github.com/doublezerofoundation/doublezero-offchain/pull/375/changes/321e23a441c4a74c6d60197a1a18dbf6edfd3ada

crates/solana-cli/src/command/shreds/publisher_rewards/distribute.rs
martinsander00: https://github.com/doublezerofoundation/doublezero-offchain/pull/375/changes/321e23a441c4a74c6d60197a1a18dbf6edfd3ada

crates/solana-cli/src/command/shreds/publisher_rewards/distribute.rs
martinsander00: https://github.com/doublezerofoundation/doublezero-offchain/pull/375/changes/321e23a441c4a74c6d60197a1a18dbf6edfd3ada

martinsander00: Addressed 1 and 2 (CLI version check hoisted out, ATAs pre-checked once). Did 3 as a cache too, but it caused mid-pass `BlockhashNotFound` in the field — the pass submits sequentially, so on a validator with many pending epochs the single cached blockhash aged out partway through and every later tx failed. Reverted to a fresh blockhash per tx (matches the rest of the codebase); the saved RPCs were not worth the correctness risk. Skipped the PDA hoist — both derivations are ~µs and live inside the SDK account builder.

bgm-malbeclabs:
### Code review

Found 1 issue:

1. The vendored Pod struct `ValidatorRewardsLeaf` has no compile-time size assertion, unlike the other mirrored structs added in this PR. Its own header comment requires it be "Kept byte-for-byte compatible with the canonical impl so merkle leaves hash identically on both sides", and it is used to compute merkle leaves/proofs that must match on-chain. A field-order or padding drift in the upstream definition would silently produce wrong proofs with no compile-time signal. The sibling mirrors `ShredDistribution` / `ShredDistributionJournal` already pin their layout with `const _: () = assert!(std::mem::size_of::<...>() == N)`. Add an equivalent `const _: () = assert!(std::mem::size_of::<ValidatorRewardsLeaf>() == 40);`.

https://github.com/doublezerofoundation/doublezero-offchain/blob/3d928c8c5c293f9144365d16a1c8d0551812338b/crates/solana-sdk/src/shred_subscription/types/validator_rewards_leaf.rs#L9-L18

Sibling structs that establish the pin pattern:

https://github.com/doublezerofoundation/doublezero-offchain/blob/3d928c8c5c293f9144365d16a1c8d0551812338b/crates/solana-sdk/src/shred_subscription/state.rs#L915-L918

martinsander00:
Good catch — added in df108cb. `const _: () = assert!(std::mem::size_of::<ValidatorRewardsLeaf>() == 40)`, mirroring the canonical `assert_eq!` and the sibling `ShredDistribution`/`ShredDistributionJournal` pins.

bgm-malbeclabs (approved):

PR #378, opened 2026-06-01
refactor(passport): extract passport verbs into doublezero-passport-cli crate

juan-malbeclabs:
## What

Extracts the `passport` command tree out of `solana-cli` into a new **library-only** crate `doublezero-passport-cli` (RFC-20 module crate). The crate exports a `Command` enum driven by `execute(ctx, out)`, reading all connection/identity config from a `CliContext` and routing output through an injected `Write` — so both the unified `doublezero` binary and the offchain `doublezero-solana` binary can mount the same verbs.

`solana-cli` keeps a thin **backward-compat adapter** (`command/passport.rs`) that re-declares the exact pre-RFC-20 flag surface (per-verb `--url`, `--keypair`, signer flags), builds a `CliContext` from them, and delegates to the library.

## Behavior parity (the important part)

The `doublezero-solana passport ...` CLI behaves **exactly as before**, even where that diverges from RFC-20:

- `fetch` keeps the legacy pipe-table output with `{:.9} SOL` precision.
- `find-validator` keeps the branch-specific gossip warnings ("Your node ID..." vs "Your IP...") and the print-and-return handling of IP parse / detection failures (exit code 0, not error propagation).
- `prepare-` / `request-validator-access` validation output is character-for-character identical.

`--json` / `--json-compact` are **additive opt-in** flags on the read verbs; any invocation without them is byte-identical to the previous CLI.

## Verification

Built two binaries from the same tree (old `HEAD` vs new working tree) and diffed stdout/stderr/exit-code against a **live RPC** (testnet + mainnet) across all verbs and their error paths — every legacy invocation is identical. `--help` surfaces match except for the additive `--json` flags and a cosmetic option-ordering change in `request-validator-access --help`.

- `cargo test -p doublezero-passport-cli -p doublezero-solana-cli`: all green (4 + 134 + 6).
- `cargo clippy`: clean.

## Cleanup

Removes the now-duplicated gossip/cluster helpers from `solana-cli/src/utils.rs` (moved into the new crate); centralizes the JSON/format-resolution logic in `passport-cli/src/output.rs`.


karl-dz (commented):
The output-parity work here looks carefully put together. I traced the migrated `writeln!`/`write!` calls against the deleted `println!`s, and the newline structure, emoji, `{:.9} SOL` precision, pipe-table headers, and branch-specific gossip warnings all line up. The `Write`-threading is done correctly throughout, including the subtle choice to hold an unlocked `std::io::stdout()` so the handle stays `Send` across awaits. The backward-compat adapter is appropriately thin, and the `command_tree_is_valid` `debug_assert` plus the legacy/JSON parse tests in `main.rs` are the right guard for a clap-surface refactor.

My agents and I found 5 issues and 3 nits on this review.

The most significant of these concerns error handling. The crate leans on both `anyhow` (internally) and `eyre` (at the boundary) and flattens between them, which discards the cause chain. As a reusable library it would be cleaner to drop both and expose typed errors (native errors where one source suffices, a `thiserror` `PassportCliError` where several combine), leaving each consumer to lift into `anyhow` or `eyre` as it prefers. Details inline.

One structural note with no single-line anchor: `Cargo.lock` now carries `doublezero-config` and `doublezero-program-common` at both 0.20.0 and 0.25.0. The `Environment` type unifies correctly at the adapter boundary (both `doublezero-cli-core` and `doublezero-config` are pinned to 0.25.0, and only the 0.25 `Environment` is ever touched), so it compiles and is correct today. It does add build time and binary size, and leaves a latent type-confusion footgun if anything later passes a 0.20 `Environment` here. Worth a follow-up to converge the workspace on `client/v0.25.0`.

crates/passport-cli/src/command.rs
```diff
@@ -0,0 +1,36 @@
...
+    /// Validate arguments and generate the required transaction signature command
+    PrepareValidatorAccess(prepare_access::PrepareValidatorAccessArgs),
+    /// Request access as a Solana Validator
+    RequestValidatorAccess(request_access::RequestValidatorAccessArgs),
+}
+
+impl Command {
+    /// Dispatch to the selected verb. All output is written to `out`; all
+    /// configuration is read from `ctx`.
+    pub async fn execute(self, ctx: &CliContext, out: &mut impl Write) -> eyre::Result<()> {
```
karl-dz: This crate is a reusable library, but it currently routes errors through two application-level error crates at once. Every verb's internal helpers return `anyhow::Result`, the public `execute` boundary returns `eyre::Result`, and the bridge flattens with `eyre::eyre!("{e:#}")`. The `solana-cli` adapter then flattens again through `into_anyhow`. Each `"{e:#}"` collapses the cause chain into a single string, so by the time an error reaches a binary the structured chain is gone, which also defeats `doublezero-cli-core`'s `render_eyre` (the helper that prints the chain of causes RFC-20 §Error handling calls for).

A library is better off owning neither `anyhow` nor `eyre` and exposing typed errors instead. The shape depends on the verb:
- Where a verb propagates a single existing error, return it directly (for example `Result<T, ClientError>`).
- Where a verb aggregates several sources (RPC fetches, signature parsing, instruction building, writer IO), define a `thiserror` enum such as `PassportCliError` that wraps them with `#[from]`.

Each consumer then lifts the typed error with a single `?`: the `solana-cli` adapter into `anyhow`, the future unified `doublezero` binary into `eyre`. Because each conversion is a real `From` rather than a string flatten, the cause chain survives in both, and the crate carries no application error dependency of its own.

RFC-20 notes that module errors conventionally flow as `eyre::Report`, but that is a convention rather than a trait constraint (dispatch is one match arm per module), so a typed error composes through `?` just as well. Either way, the `"{e:#}"` flattening should go.

crates/passport-cli/src/find_validator.rs
```diff
@@ -0,0 +1,250 @@
...
+    sol_client: &SolRpcClient,
+    out: &mut W,
+) -> Result<()> {
+    writeln!(out, "Validator ID: {}", node.pubkey)?;
+    match &node.gossip {
+        Some(gossip) => writeln!(out, "Gossip IP: {}", gossip.ip())?,
+        None => writeln!(out, "Gossip IP: <unknown>")?,
+    }
+
+    let pubkey = node.pubkey.parse::<Pubkey>().expect("Invalid pubkey");
```
karl-dz: `node.pubkey` comes straight from the RPC `getClusterNodes` response, and `.parse::<Pubkey>().expect("Invalid pubkey")` turns a malformed or buggy RPC reply into a process panic with a backtrace rather than a handled error. The same `.expect` appears in `run_json`'s `Some(node)` arm. Both `print_node_info` and `run_json` already return `Result`, so this should propagate the failure instead of panicking. `.parse::<Pubkey>()` already yields a `ParsePubkeyError`, so the minimal form is just `?`. If the crate adopts a typed error, wrap it in a variant that carries the offending string as context (for example `PassportCliError::InvalidValidatorPubkey { pubkey, source }`).

The parsed `Pubkey` is only needed for the `is_scheduled_leader` call. The lookup itself (`find_node_by_node_id` / `find_node_by_ip`) already matches on the raw string. This is moved verbatim from the deleted code, so it predates this PR, but it lives in the relocated files and is cheap to convert while they are being touched.

crates/passport-cli/src/util.rs
```diff
@@ -0,0 +1,94 @@
...
+            Cluster::Unknown => write!(f, "unknown"),
+        }
+    }
+}
+
+pub async fn identify_cluster(client: &RpcClient) -> Cluster {
+    let genesis_hash = client
+        .get_genesis_hash()
+        .await
+        .expect("Failed to fetch genesis hash");
```
karl-dz: `identify_cluster` calls `get_genesis_hash().await.expect("Failed to fetch genesis hash")`, so any transient RPC or network error becomes a process panic with a backtrace. It is called purely to print the `Connected to Solana: {cluster}` label, from `request_access`, `find_validator` (both the human and JSON paths), and `prepare_access`. A network blip on a display-only lookup should not abort the command with a panic.

`get_genesis_hash()` surfaces a `ClientError` on failure, so `identify_cluster` should return `Result<Cluster, ClientError>` (or the crate's typed error) and let the four call sites propagate with `?`. That still fails the command on a genuine RPC outage, but with a clean single-line error instead of a panic. If the label should instead degrade gracefully, returning `Cluster::Unknown` on error is the alternative, though that changes the current abort-on-failure behavior. Pre-existing and moved verbatim, but cheap to fix while the function is being relocated.

crates/passport-cli/src/find_validator.rs
```diff
@@ -0,0 +1,250 @@
...
+    /// branch-specific warnings and the print-and-return handling of parse / IP
+    /// detection failures.
+    async fn run_human(self, ctx: &CliContext, out: &mut impl Write) -> Result<()> {
+        tracing::debug!(env = %ctx.env, "passport find-validator");
+
+        writeln!(out, "DoubleZero Passport - Find Validator")?;
+
+        let connection = SolanaConnection::new(ctx.solana_l1_rpc_url.clone());
+        let sol_client = SolRpcClient::new(
+            Url::parse(&connection.url()).unwrap(),
```
karl-dz: `Url::parse(&connection.url()).unwrap()` appears at four sites: here and at line 139 in this file, plus `request_access` and `prepare_access`. The string is round-tripped from an already-constructed `SolanaConnection`, so for well-formed monikers and URLs it parses and the `unwrap` is effectively unreachable. It is still an `unwrap` on a value derived from user input (`--url`), and the enclosing functions all return `Result`, so `?` on the native `url::ParseError` (or a typed-error conversion) removes the panic path for the cost of one character. Lowest-stakes of the panic sites, and pre-existing, but worth converting alongside the others.

crates/passport-cli/src/fetch.rs
```diff
@@ -0,0 +1,188 @@
...
+struct AccessRequestView {
+    access_request: String,
+    service_key: String,
+    rent_beneficiary_key: String,
+    request_fee_sol: f64,
+    access_mode: String,
+}
+
+impl FetchArgs {
+    pub async fn execute(self, ctx: &CliContext, out: &mut impl Write) -> eyre::Result<()> {
```
karl-dz: Since this PR is about preserving byte-for-byte output parity with the pre-RFC-20 CLI, it would help to have that parity guarded in-tree. Right now it rests on the manual old-vs-new binary diff described in the PR body, and none of the human-output paths have tests, so a future edit to any `writeln!` could break parity with nothing in CI to catch it.

Because every verb now takes `out: &mut impl Write`, these are trivial to exercise against an in-memory buffer. The `access_validation` tests already do exactly this with `let mut out = Vec::new();`. Worth adding golden-output assertions for at least:
- `fetch --config` (the pipe-table layout and the `{:.9} SOL` precision), and
- the `find_validator` not-in-gossip warning branches (the "Your node ID..." vs "Your IP..." wording).

Of the follow-ups here, this is the one most worth doing given the PR's goal.

crates/passport-cli/src/output.rs
```diff
@@ -0,0 +1,41 @@
...
+//! keep the format-resolution and JSON-emission logic in one place.
+
+use std::io::Write;
+
+use doublezero_cli_core::OutputFormat;
+use serde::Serialize;
+
+/// Honor the per-verb `--json` / `--json-compact` flags when set, otherwise fall
+/// back to the context's output-format hint supplied by the binary.
+pub fn resolve_format(json: bool, json_compact: bool, ctx_format: OutputFormat) -> OutputFormat {
```
karl-dz: Nit: `--json` / `--json-compact` are declared on the verb structs (`FetchArgs`, `FindValidatorArgs`), and the adapter flattens those in via `inner`. So `build_ctx` reads `inner.json`/`inner.json_compact` into `ctx.output_format`, and then the verb reads `self.json`/`self.json_compact` again and reconciles the two via `resolve_format`. The same two booleans reach the same decision by two paths. Would it be cleaner for the verbs to read format only from `ctx.output_format`, and move the additive `--json`/`--json-compact` declarations onto the adapter structs (alongside `keypair_path`) where they feed `build_ctx`? That gives one source of truth (the unified binary sets it via its global output flag, the legacy adapter via its per-verb flag) and lets `resolve_format` go away.

crates/passport-cli/src/find_validator.rs
```diff
@@ -0,0 +1,250 @@
...
+                Err(e) => writeln!(out, "Failed to get public IP: {e}")?,
+            }
+        }
+
+        Ok(())
+    }
+
+    /// Additive JSON output for the read verb. Collects the same lookup into a
+    /// serializable view.
+    async fn run_json(
```
karl-dz, line 97: Nit: `run_human` and `run_json` each re-implement the same three-way node lookup (by `--validator-id`, by `--gossip-ip`, else by detected public IP), and the `is_scheduled_leader` call is duplicated between `print_node_info` and the `Some(node)` arm of `run_json`. The additive JSON path made this awkward to avoid, but it landed as a parallel flow rather than a shared resolver. Would extracting one `resolve_node(&self, &nodes)` that returns the matched node plus the detected IP let both paths render a single result and drop one copy of the ladder?

crates/passport-cli/src/fetch.rs
```diff
@@ -0,0 +1,188 @@
...
+                    is_request_access_paused: program_config.is_request_access_paused(),
+                    admin_key: program_config.admin_key.to_string(),
+                    sentinel_key: program_config.sentinel_key.to_string(),
+                    request_deposit_sol: program_config.request_deposit_lamports as f64 * 1e-9,
+                    request_fee_sol: program_config.request_fee_lamports as f64 * 1e-9,
+                    solana_validator_backup_ids_limit: program_config
+                        .solana_validator_backup_ids_limit
+                        as u64,
+                };
+                emit_json(out, &view, format)?;
```
karl-dz: Nit: with `--json` (or `--json-compact`) and both `--config` and `--access-request` set, `emit_json` is called twice independently, so the output is two top-level JSON objects back-to-back rather than one document. The human table path was always sequential, so this matched before, but a JSON consumer piping to `jq` will choke on the concatenation. Since `--json` is new here, would it be worth emitting a single combined object (e.g. `{ "program_config": ..., "access_request": ... }`) when both are requested?

crates/passport-cli/src/command.rs
juan-malbeclabs: Done in c7552ba. Added a typed `PassportCliError` (thiserror); the crate no longer depends on `anyhow` or `eyre`. Native single-source errors are `#[from]` variants (`ClientError`, `ParsePubkeyError`, `url::ParseError`, etc.), and the dependencies that only expose `anyhow::Error` (`Wallet`, `try_fetch_zero_copy_data`, `try_build_instruction`) are boxed through a transparent `Other(Box<dyn Error + Send + Sync>)` so the cause chain survives. Both the `eyre::eyre!("{e:#}")` flattens and the adapter's `into_anyhow` are gone — the solana-cli adapter now lifts with a plain `?` into `anyhow`, and a unified eyre binary can do the same.

crates/passport-cli/src/find_validator.rs
juan-malbeclabs: Fixed. Both sites now go through a shared `leader_status()` helper that does `node.pubkey.parse::<Pubkey>()?`, surfacing a `ParsePubkeyError` instead of panicking.

crates/passport-cli/src/util.rs
juan-malbeclabs: Fixed. `identify_cluster` now returns `Result<Cluster, PassportCliError>` and propagates the `ClientError` from `get_genesis_hash`; the four call sites use `?`. Kept the abort-on-failure behavior (no silent `Unknown`), just without the panic.

crates/passport-cli/src/find_validator.rs
juan-malbeclabs: Fixed at all four sites — now `?` on the native `url::ParseError`.

crates/passport-cli/src/fetch.rs
juan-malbeclabs: Added. Extracted the human renderers into pure functions and added in-tree assertions: `fetch --config` (full pipe-table layout + `{:.9} SOL` precision, including a 1-lamport case) and the find-validator not-in-gossip branches ("Your node ID…" vs "Your IP…").

crates/passport-cli/src/output.rs
juan-malbeclabs: Done. The `--json`/`--json-compact` flags now live on the solana-cli adapter structs (next to the connection options) and feed `build_ctx`; the verbs read format only from `ctx.output_format`. `resolve_format` is gone. CLI surface is unchanged (verified the flags and their mutual exclusivity still work).

crates/passport-cli/src/find_validator.rs
juan-malbeclabs: Partially. Extracted the duplicated `is_scheduled_leader` call into `leader_status()`, and pulled the lookup+render tails into `render_node_id_node` / `render_ip_node`. I deliberately did not merge the human and JSON ladders into one resolver: the human path's graceful "print a message and return Ok" on parse / IP-detection failures (vs. the JSON path propagating) is load-bearing for the byte-for-byte parity this PR targets, and a single resolver would blur that.

crates/passport-cli/src/fetch.rs
juan-malbeclabs: Done. When both `--config` and `--access-request` are requested with JSON, the verb now emits a single `{ "program_config": …, "access_request": … }` object; a single section still serializes flat.

karl-dz:
@juan-malbeclabs Fix for the CI issue just landed. Can you rebase vs origin/main?

karl-dz (approved):

PR #379, opened 2026-06-02
chore(deps): converge SDK family on client/v0.25.1

juan-malbeclabs:
## Why

The `Cargo.lock` carried **both** `v0.20` and `v0.25` of `doublezero-config` and `doublezero-program-common`: `doublezero-cli-core` / `doublezero-config` were pinned to `client/v0.25.1` while `doublezero_sdk` and the rest of the SDK family stayed at `client/v0.20.0` and dragged the `v0.20` copies in transitively. This is the follow-up tracked out of PR #378's "Cargo.lock 0.20/0.25 convergence" note.

## What

Bump the remaining SDK family — `doublezero_sdk`, `-serviceability`, `-record`, `-telemetry`, `-program-common` — from `client/v0.20.0` to `client/v0.25.1` so the whole family resolves to a single version. The duplicate `v0.20` entries drop out of the lockfile (net **-23 lines** in `Cargo.lock`).

The `v0.20 → v0.25` serviceability jump changes a few account layouts; the only crate that broke was `contributor-rewards`. Fixes, all mirroring the SDK 0.25 source:

- **`Device.interfaces`** is now a flat `Vec<Interface>` (the versioned `InterfaceDeprecated` enum moved to `Device.deprecated_interfaces`) → drop the obsolete `Interface::into_current_version()` projection in the Shapley handler.
- **`UserStatus::Rejected`/`PendingBan`** were renamed to `*Deprecated` → update the liveness filter in the demand ingestor (and its mirror in the test), **preserving the exact set of non-live statuses**.
- **`apply_json_compat_migrations`** extended so pre-`v0.25` snapshots still load: seed `Device.deprecated_interfaces` from the legacy enum vec, project `Device.interfaces` onto the flat struct (mirroring `TryFrom<&InterfaceV1/V2> for Interface` defaults), and backfill the new `User.bgp_rtt_ns` field.

## Scope note — `UserStatus::OutOfCredits`

`UserStatus` in 0.25 adds a variant that didn't exist in 0.20: `OutOfCredits`. This PR is deliberately **behavior-neutral** — only the three original statuses (`RejectedDeprecated | Banned | PendingBanDeprecated`) count as non-live, so an `OutOfCredits` multicast subscriber is still counted as active demand. Whether `OutOfCredits` should be treated as non-live is a reward-math/product decision, intentionally left out of this convergence. Flagging for a separate call.

## Testing

- `cargo test --workspace` — 43 test binaries pass
- `cargo clippy --workspace --all-targets` — clean

### DIFF
--- a/crates/contributor-rewards/src/ingestor/types.rs
+++ b/crates/contributor-rewards/src/ingestor/types.rs
@@ -7,9 +7,14 @@ use anyhow::{Result, bail};
 use chrono::{DateTime, Utc};
 use doublezero_program_common::serializer;
 use doublezero_serviceability::state::{
-    accesspass::AccessPass as DZAccessPass, contributor::Contributor as DZContributor,
-    device::Device as DZDevice, exchange::Exchange as DZExchange, link::Link as DZLink,
-    location::Location as DZLocation, multicastgroup::MulticastGroup as DZMulticastGroup,
+    accesspass::AccessPass as DZAccessPass,
+    contributor::Contributor as DZContributor,
+    device::Device as DZDevice,
+    exchange::Exchange as DZExchange,
+    interface::{CURRENT_INTERFACE_SCHEMA_VERSION, INTERFACE_MTU},
+    link::Link as DZLink,
+    location::Location as DZLocation,
+    multicastgroup::MulticastGroup as DZMulticastGroup,
     user::User as DZUser,
 };
 use doublezero_telemetry::state::{
@@ -123,7 +128,9 @@ fn apply_serviceability_json_compat_migrations(serviceability: &mut Value) {
         }
     }
 
-    // doublezero-serviceability v0.20 added durable tunnel/BGP state to User.
+    // doublezero-serviceability v0.20 added durable tunnel/BGP state to User;
+    // client/v0.25.0 then appended `bgp_rtt_ns`. Both default to the same values
+    // onchain/Borsh deserialization uses for absent tails.
     if let Some(users) = serviceability
         .get_mut("users")
         .and_then(|users| users.as_object_mut())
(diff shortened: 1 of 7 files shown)


crates/contributor-rewards/src/ingestor/types.rs
```diff
@@ -139,10 +146,111 @@ fn apply_serviceability_json_compat_migrations(serviceability: &mut Value) {
...
+/// `{"V2": {...}}`) onto the flat `Interface` struct introduced in
+/// doublezero-serviceability client/v0.25.0. Mirrors the defaults stamped by the
+/// SDK's `TryFrom<&InterfaceV1>` / `TryFrom<&InterfaceV2> for Interface` impls:
+/// a V1 body fans in through V2, gaining `interface_cyoa`/`interface_dia` =
+/// `None`, `bandwidth`/`cir` = 0, `mtu` = `INTERFACE_MTU`, `routing_mode` =
+/// `Static`; both versions gain `version` = `CURRENT_INTERFACE_SCHEMA_VERSION`
+/// and an empty `flex_algo_node_segments`. `size` is the on-disk byte length,
+/// which offchain consumers never read (only `interface_type`/`bandwidth` are
+/// used), so it is defaulted to 0 rather than recomputed.
+fn migrate_interface_to_flat(interface: &mut Value) {
```
vihu, line 201: I think this function does a bit more than probably necessary? This likely accomplishes the same thing:

```rust
fn migrate_interface_to_flat(interface: &mut Value) {
    let Some(obj) = interface.as_object() else {
        return;
    };
    // Already in the flat v0.25 form; nothing to do.
    if obj.contains_key("size") {
        return;
    }

    // Pull out the versioned body and note whether it's V1, which predates the
    // CYOA/DIA/bandwidth/routing fields and so needs them backfilled.
    let (is_v1, body) = if let Some(body) = obj.get("V1").and_then(Value::as_object) {
        (true, body)
    } else if let Some(body) = obj.get("V2").and_then(Value::as_object) {
        (false, body)
    } else {
        return;
    };

    let mut flat = body.clone();

    if is_v1 {
        flat.entry("interface_cyoa")
            .or_insert_with(|| Value::String("None".into()));
        flat.entry("interface_dia")
            .or_insert_with(|| Value::String("None".into()));
        flat.entry("bandwidth")
            .or_insert_with(|| Value::Number(0.into()));
        flat.entry("cir").or_insert_with(|| Value::Number(0.into()));
        flat.entry("mtu")
            .or_insert_with(|| Value::Number(INTERFACE_MTU.into()));
        flat.entry("routing_mode")
            .or_insert_with(|| Value::String("Static".into()));
    }

    flat.entry("flex_algo_node_segments")
        .or_insert_with(|| Value::Array(Vec::new()));
    flat.insert(
        "version".into(),
        Value::Number(CURRENT_INTERFACE_SCHEMA_VERSION.into()),
    );
    flat.insert("size".into(), Value::Number(0.into()));

    *interface = Value::Object(flat);
}
```

vihu (approved):

PR #381, opened 2026-06-04
solana-cli: publisher-rewards per-epoch status command

martinsander00:
## Summary
- Publishers had no way to see their shred rewards by epoch. Adds `doublezero-solana shreds publisher-rewards status --node-id <PUBKEY> [--num-epochs N]`, a read-only per-epoch table keyed by node identity.
- Each epoch with an on-chain `ShredDistribution` always gets a row with a clear status — `not ready` / `ready` / `claimed` / `no rewards` (validator wasn't a publisher) / `no data` (export not available yet) — so nothing is ever silently dropped.
- Shows the reward token (MINT) and amount per epoch: `claimed` amounts are read exactly from the distribute transaction's destination-ATA balance delta; `ready` amounts are an estimate (`~`) projected from the on-chain pool/slots/proportion/burn math.
- Reuses the distribute scan (S3 leaf fan-out + journal/bitmap reads) so the table reflects exactly what a distribute pass would act on; `distribute::bitmap_bit_set` is now `pub(crate)` to share the single decision predicate.

## Testing Verification
- Unit tests covering the hand-ported `client_proportion` lookup (default / legacy-35% fallback / per-client override gated by `set_bitmap`) and the amount formatting.
- Manually exercised against mainnet for a live validator across `ready` / `claimed` / `no rewards` / `no data` epochs.

## Example Output
```
Node ID:       xxx
Rewards owner: xxx
Rewards mint:  2Z

Scanning epochs 961..=982.

| EPOCH | LEADER SLOTS | MINT | AMOUNT | STATUS    |
|-------|--------------|------|--------|-----------|
| 961   |          104 | 2Z   |   10.9 | claimed   |
| 962   |          104 | 2Z   |   10.1 | claimed   |
| 963   |          120 | 2Z   |     14 | claimed   |
| 964   |          164 | 2Z   |   20.6 | claimed   |
| 965   |          112 | 2Z   |     13 | claimed   |
| 966   |          120 | 2Z   |   16.2 | claimed   |
| 967   |          152 | 2Z   |   18.5 | claimed   |
| 968   |          140 | 2Z   |  ~21.3 | ready     |
| 969   |          124 | 2Z   |  ~18.4 | ready     |
| 970   |          132 | 2Z   |    ~22 | ready     |
| 971   |          148 | 2Z   |  ~23.8 | ready     |
| 972   |          172 | 2Z   |  ~26.1 | ready     |
| 973   |          132 | 2Z   |  ~18.3 | ready     |
| 974   |          156 | —    |      — | not ready |
| 975   |          148 | —    |      — | not ready |
| 976   |          196 | —    |      — | not ready |
| 977   |          112 | —    |      — | not ready |
| 978   |          124 | —    |      — | not ready |
| 979   |          120 | —    |      — | not ready |
| 980   |          132 | —    |      — | not ready |
| 981   |          136 | —    |      — | not ready |
| 982   |            — | —    |      — | no data   |

Status:
  ready       reward is accumulated and waiting to be distributed to your ATA
  not ready   reward for this epoch isn't distributable yet (still being collected, calculated, or swapped)
  claimed     reward has already been distributed to your ATA
  no rewards  you published no shreds this epoch, so there's nothing to distribute
  no data     this epoch's leader-slot export isn't available yet
```

### DIFF
--- a/crates/solana-cli/src/command/shreds/publisher_rewards/status.rs
+++ b/crates/solana-cli/src/command/shreds/publisher_rewards/status.rs
@@ -192,7 +192,7 @@ impl StatusCommand {
             .await
             .context("fetching current Solana epoch")?
             .epoch;
-        let from_epoch = current_epoch.saturating_sub(self.num_epochs + 1);
+        let from_epoch = current_epoch.saturating_sub(self.num_epochs);
 
         println!("\nScanning epochs {from_epoch}..={current_epoch}.");
 
@@ -497,8 +497,10 @@ async fn fetch_mint_decimals(
         .collect()
 }
 
-/// Format a raw token amount to at most one (truncated) fractional digit,
-/// falling back to the raw integer when decimals are unknown.
+/// Format a raw token amount to at most three (truncated) fractional digits with
+/// trailing zeros trimmed, falling back to the raw integer when decimals are
+/// unknown. Keeping three digits stops small-but-real payouts (e.g. 0.05 SOL at
+/// 9 decimals) from collapsing to `0`.
 fn format_amount(raw: u64, decimals_by_mint: &HashMap<Pubkey, u8>, mint: &Pubkey) -> String {
     let Some(&decimals) = decimals_by_mint.get(mint) else {
         return raw.to_string();
@@ -509,12 +511,13 @@ fn format_amount(raw: u64, decimals_by_mint: &HashMap<Pubkey, u8>, mint: &Pubkey
     let scale = 10u128.pow(decimals as u32);
     let raw = raw as u128;
     let integer = raw / scale;
-    let tenth = (raw % scale) * 10 / scale;
-    if tenth == 0 {
-        integer.to_string()
-    } else {
-        format!("{integer}.{tenth}")
+    // Truncate the fraction to three digits, then drop trailing zeros.
+    let frac = (raw % scale) * 1_000 / scale;
+    if frac == 0 {
+        return integer.to_string();
     }
+    let frac = format!("{frac:03}");
+    format!("{integer}.{}", frac.trim_end_matches('0'))
 }
 
 /// Resolve `(epoch → (reward mint, exact paid amount))` for `claimed` epochs by
@@ -627,7 +630,13 @@ async fn resolve_distributed_payouts(
             .unwrap_or_default();
         let paid = post_raw.saturating_sub(pre_raw) as u64;
         if let Ok(mint) = destination.mint.parse::<Pubkey>() {
-            resolved.entry(epoch).or_insert((mint, paid));
+            // A multi-client validator has multiple leaves per epoch, and
+            // distribute emits one tx per leaf crediting the same ATA, so sum
+            // them to match `estimate_publisher_payout`'s per-leaf total.
+            resolved
+                .entry(epoch)
+                .and_modify(|(_, amount)| *amount += paid)
+                .or_insert((mint, paid));
         }
     }
     resolved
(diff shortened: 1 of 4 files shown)


crates/solana-cli/src/command/shreds/publisher_rewards/status.rs
```diff
@@ -0,0 +1,836 @@
...
+    journals: &[solana_sdk::account::Account],
+    parent_accounts: &[solana_sdk::account::Account],
+    parent_index_for_dz_epoch: &HashMap<u64, usize>,
+) -> (RewardStatus, Option<Pubkey>, AmountInfo) {
+    // Distribute won't pay until the parent distribution is finalized.
+    let parent = parent_index_for_dz_epoch
+        .get(&epoch.associated_dz_epoch)
+        .and_then(|&index| parent_accounts.get(index))
+        .filter(|account| !account.data.is_empty())
+        .and_then(ZeroCopyAccountOwnedData::<ParentDistribution>::from_account)
```
bgm-malbeclabs, line 722: **Claimed amount underreported for multi-`client_id` validators.**

`estimate_publisher_payout` **sums** the payout across all of the validator's leaves, but here the claimed amount uses `or_insert`, which keeps only the first resolved tx for the epoch.

A validator running multiple clients has multiple leaves in one epoch, and distribute emits one ix per leaf, one tx each (`distribute.rs:507-545`), so each leaf's payout lands in a separate transaction crediting the same destination ATA. `or_insert` therefore keeps only one leaf's amount — and since the txs are gathered via `buffer_unordered`, *which* one is nondeterministic. Result: a multi-leaf validator shows the full sum while `ready` (`~`), but only one leaf's amount once `claimed`.

Suggest accumulating instead:
```rust
resolved
    .entry(epoch)
    .and_modify(|(_, a)| *a += paid)
    .or_insert((mint, paid));
```
Single-client validators are unaffected, so this is easy to miss in testing.

crates/solana-cli/src/command/shreds/publisher_rewards/status.rs
```diff
@@ -0,0 +1,836 @@
...
+                ShredSubscriptionInstructionData::try_from_slice(&ix.data),
+                Ok(ShredSubscriptionInstructionData::DistributeValidatorRewards { .. })
+            ) {
+                continue;
+            }
+            for &account_index in &ix.accounts {
+                if let Some(epoch_for_key) = account_keys
+                    .get(account_index as usize)
+                    .and_then(|key| pda_to_epoch.get(key))
+                {
```
bgm-malbeclabs, line 604: **Rewards below 0.1 token render as `0`.**

Truncating to a single fractional digit means any nonzero amount under 0.1 of a token displays as `0`. With WSOL at 9 decimals, 0.05 SOL → `tenth == 0` → `"0"`; same for sub-0.1 USDC/2Z. A `claimed 0` / `ready ~0` for an actually-nonzero payout is indistinguishable from `no rewards`, which is misleading for a read-only status command. Consider more fractional digits (e.g. trim trailing zeros at ~4 places) so small-but-real amounts stay visible.

crates/solana-cli/src/command/shreds/publisher_rewards/status.rs
```diff
@@ -0,0 +1,836 @@
...
+
+    if existing.is_empty() {
+        return Ok(Vec::new());
+    }
+
+    // Fan out S3 to find this validator's leaves per epoch.
+    let s3_client = s3::build_s3_client()?;
+    let mut fetch_results = stream::iter(existing.iter())
+        .map(|(epoch, shred_distribution)| {
+            let s3_client = s3_client.clone();
```
bgm-malbeclabs, line 278: **Window scans `num_epochs + 2` epochs, not the documented count.**

`from_epoch = current_epoch.saturating_sub(self.num_epochs + 1)` scanned inclusively over `from_epoch..=current_epoch` is `num_epochs + 2` epochs — with the default `--num-epochs 20` that's `current-21..=current` = 22 epochs, but the changelog and the `--num-epochs` help text both say "20". Minor (the printed `Scanning epochs X..=Y` is accurate and it just scans a couple extra), but the count doesn't match the stated semantics. Dropping the `+ 1` makes it match; worth deciding whether `N` means "N rows" or "N back from current".

bgm-malbeclabs (commented):
Reviewed the new `status.rs`. The core reward math and the ready/claimed bitmap inference check out against the on-chain `DistributeValidatorRewards` handler (same `leader_slots * (MAX − proportion)` scaling, denominator, `rewards_amount()` selection, 3500 legacy fallback; no under/overflow risk since `mul_scalar(x) ≤ x` and `client_proportion ≤ UnitShare16::MAX`). Bit semantics and `bitmap_bit_set` ordering also match.

Three things worth a look — inline comments below. #1 is a real correctness bug for multi-`client_id` validators.

crates/solana-cli/src/command/shreds/publisher_rewards/status.rs
martinsander00: https://github.com/doublezerofoundation/doublezero-offchain/pull/381/changes/9770a1c58e2a6881f724f0449409ff4c0187107a

crates/solana-cli/src/command/shreds/publisher_rewards/status.rs
martinsander00: https://github.com/doublezerofoundation/doublezero-offchain/pull/381/changes/9770a1c58e2a6881f724f0449409ff4c0187107a

crates/solana-cli/src/command/shreds/publisher_rewards/status.rs
martinsander00: https://github.com/doublezerofoundation/doublezero-offchain/pull/381/changes/9770a1c58e2a6881f724f0449409ff4c0187107a

bgm-malbeclabs (approved):

PR #383, opened 2026-06-05
refactor(solana-cli): RFC-20 verb contract for all CLI verbs

ben-dz:
## Reviewing this PR

History is organized into **3 commits by area** — review by commit:

1. **core plumbing + passport** — the RFC-20 `execute(&ctx, …)` contract, the `CliContext` + global-flags wiring in `main.rs`, the per-verb→ctx merge helpers (`build_wallet`, `solana_connection`, `resolve_network_env`), payer plumbing, and the passport verbs as the exemplar. The behavior-relevant commit — review closely.
2. **shreds verbs** — convert to the new signature and route each verb's per-verb connection/keypair override through the ctx merge helpers. Also includes the `publisher-rewards status` verb (landed on main in #381) converted to the contract, and the `shreds --dz-ledger-url` subcommand-position flag retained as before.
3. **revenue-distribution verbs** — same conversion (incl. the in-place `relay` subtree, which keeps `SolanaPayerOptions` for the `Schedulable` trait pending #1520).

Note: commits 1–2 don't individually compile by design (the dispatch lands in commit 1; the verbs it calls are converted in 2–3) — review cumulatively; the PR head is what builds.

**Rebased onto Solana v3 `main`** (malbeclabs/doublezero#3830 + offchain #397). Because `main` had independently rewritten several shreds verbs, `claim`, `list`, `withdraw`, and `status` were re-derived from main's current v3 bodies with the RFC-20 transform applied (not the pre-v3 bodies). Connections resolve to the network's Solana L1 for all monikers, matching #396's removal of the testnet DZ-Ledger special-case (reach the testnet program with `-u devnet`).

---
Parent Issue for Migration Process: https://github.com/malbeclabs/infra/issues/1515

## Summary of Changes
* Refactor all CLI verbs under `crates/solana-cli/src/command/` to follow the RFC-20 verb contract: each verb is now `async fn execute(self, ctx: &CliContext, out: &mut impl Write) -> Result<()>`.
* Add global flags (`--env`, `--solana-url`/`--url`, `--dz-ledger-url`, `--keypair`) to the `doublezero-solana` binary, constructing a `CliContext` at startup via `CliContextBuilder`.
* **Backwards compatible — no behavior change without the new global flags.** Per-verb `--url`/`-u`, `--keypair`/`-k`, `--dz-env`, and `shreds --dz-ledger-url` keep their positions and their meaning, including everything the per-verb moniker implies (network environment, token mints, oracle key, serviceability program id, DZ Ledger URL) — resolved per-verb from the moniker or the connection's genesis hash exactly as before (`resolve_network_env`). The global flags supply defaults; per-verb flags win when present.
* **Global `--env` uses the DoubleZero environment taxonomy** (`mainnet-beta`, `testnet`, `devnet`, `local`), matching the `doublezero` CLI these verbs will mount into (#1520): `--env devnet` selects the DZ devnet environment (Solana L1 = testnet), with every unset context field derived from `env.config()`. The Solana devnet cluster (e.g. the testnet shred-subscription program per #1763) stays reachable with a trailing `-u devnet`.
* `WriteVerbOptions` carries each write verb's per-verb connection, keypair, and signer flags; `build_wallet` merges those over the `CliContext`. Read verbs keep a per-verb `SolanaConnectionOptions` merged the same way (`solana_connection`). Passport and the relay subtree keep their existing per-verb options unchanged.
* Replace `println!` with `writeln!(out, ...)` across all verbs for testability with `Vec<u8>` writers.
* Add `Wallet::write_verbose_output` writer-aware variant in `solana-client-tools`.
* This unblocks malbeclabs/infra#1520 (mount into unified `doublezero` binary).
* Fixes malbeclabs/infra#1517

## Testing Verification
* All 161 tests pass (145 lib + 16 binary) on Solana v3, including the trailing-flag back-compat tests (`-u`/`-k`/`--dz-env` after the subcommand on read, write, and shreds verbs), the `merge_ctx` consistency tests, and the `--env` taxonomy tests; `test_command_tree_is_valid` confirms no clap argument collisions.
* `cargo clippy --all-targets` clean (no warnings); `cargo fmt` applied.
* Architecture and security reviews posted to malbeclabs/infra#1517.


crates/solana-cli/src/command/passport.rs
```diff
@@ -94,71 +94,90 @@ pub struct RequestAdapter {
...
+    // than overriding the parent, so any field added to `CliContext` is
+    // silently dropped here until mirrored below. Keep in sync, or replace with
+    // a `CliContext::with_overrides()` helper in cli-core if one lands.
     let mut builder = CliContextBuilder::new()
         .with_env(env)
         .with_solana_l1_rpc_url(solana_l1_rpc_url)
-        .with_output_format(OutputFormat::from_flags(json, json_compact));
-    if let Some(path) = keypair_path {
-        builder = builder.with_keypair_path(PathBuf::from(path));
+        .with_ledger_rpc_url(parent.ledger_rpc_url.clone())
```
juan-malbeclabs: Latent hazard (you already flag it above): `env` can be overridden by the per-verb moniker, but `ledger_rpc_url` is pinned to `parent` here and `serviceability_program_id` isn't set — so the builder derives it from the *overridden* env. That leaves those two fields inconsistent with each other for a trailing `-u <moniker>` that differs from the global `--env`.

Harmless today because passport verbs only read `solana_l1_rpc_url` / `output_format` / `env`, but it becomes a real bug the moment a verb mounted here reads the ledger or serviceability id. Deriving the ledger from the resolved env when `conn.moniker_env().is_some()` (or the `CliContext::with_overrides()` helper you mention) would close it.

crates/solana-cli/src/command/shreds/publisher_rewards/prepare_offchain_message.rs
```diff
@@ -51,20 +52,23 @@ pub struct PrepareOffchainMessageCommand {
...
             bail!("--rewards-token-owner must not be the default pubkey");
         }
 
-        let connection = SolanaConnection::from(self.connection_options);
-        let rewards_token_mint = self.rewards_token_mint.resolve(&connection).await?;
+        let solana_connection = crate::command::l1_connection(ctx, self.connection_options.clone());
+        let rewards_token_mint = self.rewards_token_mint.resolve(&solana_connection).await?;
+
+        let connection =
+            crate::command::shred_subscription_connection(ctx, &self.connection_options);
```
juan-malbeclabs: Post-v3, `l1_connection` and `shred_subscription_connection` have identical bodies, so this builds two equivalent `SolanaConnection`s per invocation (same in publisher-rewards `configure`). Reusing the one from line 67 would avoid the redundant construction. Minor.

crates/solana-cli/src/main.rs
```diff
@@ -1,13 +1,35 @@
...
+    NetworkEnvironment, SolanaConnection, SolanaConnectionOptions,
+};
 use tracing_subscriber::{EnvFilter, layer::SubscriberExt, util::SubscriberInitExt};
 
 #[derive(Debug, Parser)]
 #[command(term_width = 0)]
 #[command(version = option_env!("BUILD_VERSION").unwrap_or(env!("CARGO_PKG_VERSION")))]
 #[command(about = "DoubleZero Solana-related Commands", long_about = None)]
 struct DoubleZeroSolanaApp {
+    /// Network environment: mainnet-beta (default), testnet, localhost.
```
juan-malbeclabs: Help text is stale: it omits the `devnet` variant and labels `Localnet` as `localhost`. Since this is the headline new flag, worth listing all four values accurately.

juan-malbeclabs (dismissed):
## RFC-20 verb contract — LGTM ✅

Solid, well-scoped refactor. Every verb moves to `execute(self, ctx: &CliContext, out: &mut impl Write)`, `println!` → `writeln!(out, …)` is applied consistently, error propagation via `?` is correct, and the `Wallet::write_verbose_output` split (writer-aware core + stdout wrapper) is a clean way to keep the public API. The per-verb→ctx merge helpers preserve trailing-flag back-compat, and the parse-level tests plus `command_tree_is_valid` give good confidence there are no clap collisions. Approving.

### What I checked closely

- **Passport back-compat (`-u <moniker>` after the verb).** `merge_ctx` inherits `parent.ledger_rpc_url` while `env` can be overridden by the moniker, which leaves `ledger_rpc_url` / `serviceability_program_id` inconsistent with the resolved env. I traced every passport verb (`fetch`, `find-validator`, `prepare-validator-access`, `request-validator-access`) and they read only `ctx.solana_l1_rpc_url`, `ctx.output_format`, and `ctx.env` (debug log). None touch the ledger or serviceability program id, so this is a **latent** hazard, not an observable regression today — and the code already flags it. Left an inline note so it isn't forgotten when the unified binary starts exercising more of the context. See inline comment.

- **Env / URL decoupling.** Replacing `try_network_environment()` (RPC genesis-hash detection) with `network_env(ctx)` from `--env` is the one real behavior change: a raw `-u <testnet-url>` no longer implies testnet mints/program-ids unless `--env testnet` is also passed. It's intentional and documented in the CHANGELOG, and the fork scripts were updated to match. Worth calling out prominently in operator-facing release notes, since the failure mode (mainnet mint against a testnet RPC) is wrong-but-silent rather than an error.

### Non-blocking nits (inline)

- Redundant double `SolanaConnection` construction in `prepare-offchain-message` / publisher-rewards `configure` now that `l1_connection` and `shred_subscription_connection` resolve identically post-v3.
- `--env` help text is missing `devnet` and says `localhost` for the `Localnet` variant.
- `.map_err(|e| anyhow!("{e:#}"))` in `merge_ctx` / `build_cli_context` flattens the cause chain and uses the `anyhow!` macro; prefer `.map_err(anyhow::Error::msg)` per the repo's error-handling convention (pre-existing, just carried forward).

### Suggested follow-up

The merge helpers (`build_wallet_with_connection`'s connection-override branch, `l1_connection`/`shred_subscription_connection` per-verb-wins, and `merge_ctx`'s field-by-field rebuild) have no direct unit tests — only parse-level coverage. A small `merge_ctx` test asserting all resolved fields for a moniker-override case would catch exactly the field-drop class the inline comment describes. Not required for this PR.

crates/solana-cli/src/command/mod.rs
```diff
@@ -4,8 +4,134 @@ mod shreds;
...
+            verbose: opts.verbose,
+            fee_payer_path: opts.fee_payer_path,
+            dry_run: opts.dry_run,
+        },
+    };
+    Wallet::try_new(payer_opts, connection)
+}
+
+/// Map `CliContext.env` → `NetworkEnvironment` used by offchain crates.
+pub(crate) fn network_env(ctx: &CliContext) -> NetworkEnvironment {
```
nikw9944: Trailing per-verb `-u <moniker/url>` no longer drives env-derived values. Pre-PR, verbs derived `NetworkEnvironment` from the per-verb connection (moniker, else genesis-hash sniffing via `try_network_environment()`); now `network_env(ctx)`, `ctx.ledger_rpc_url`, and `ctx.serviceability_program_id` come only from global `--env` (default mainnet-beta). A legacy `doublezero-solana shreds pay -u devnet --device-code XYZ` now uses the mainnet USDC mint (pay.rs:341), mainnet oracle key, and mainnet DZ Ledger/serviceability id while connecting to devnet — breaking the exact workflow PR #396 shipped ("reach the testnet program with `-u devnet`"). Affected: shreds pay/withdraw/list/price/payments, revenue-distribution fetch config / contributor-rewards recipients / distribution / validator-debts, validator-deposit --fund-outstanding-debt. This contradicts the PR's back-compat claim and is inconsistent with the passport adapter, whose merge_ctx lets the per-verb moniker win for env (passport.rs:152-155). Either thread the per-verb moniker through these helpers the way passport does, or document the decoupling as a breaking change in the CHANGELOG.

crates/solana-cli/src/command/mod.rs
```diff
@@ -4,8 +4,134 @@ mod shreds;
...
+    match ctx.env {
+        Environment::MainnetBeta => NetworkEnvironment::MainnetBeta,
+        Environment::Testnet => NetworkEnvironment::Testnet,
+        Environment::Devnet => NetworkEnvironment::Devnet,
+        Environment::Local => NetworkEnvironment::Localnet,
+    }
+}
+
+/// Map `NetworkEnvironment` → `doublezero_config::Environment`.
+pub fn map_env(n: NetworkEnvironment) -> Environment {
```
nikw9944: The devnet↔devnet mapping conflates two taxonomies. `NetworkEnvironment::Devnet` is the Solana devnet cluster (api.devnet.solana.com); `Environment::Devnet` is the DZ devnet environment, which per RFC-20 lives on Solana testnet (`Environment::Devnet.config().solana_l1_rpc_url` = api.testnet.solana.com — the config crate comments this explicitly). Consequences: (1) `--env devnet` builds a self-contradictory CliContext — main.rs sets the L1 from `NetworkEnvironment::Devnet.solana_public_url()` (devnet) while ctx.env's config-defined L1 is testnet; mainnet/testnet/local agree, only devnet diverges. (2) Device-code resolution flips serviceability programs: the removed `serviceability_program_id(Devnet)` returned the testnet id DZtnuQ839pSaDMFG5q1ad2V95G82S5EC4RrB3Ndw2Heb, but `ctx.serviceability_program_id` under `--env devnet` is the DZ devnet id GYhQDKuESrasNZGyhMJhGYFtbzNijYhcrN9poSqCQVah — so even the intended replacement for the pre-PR `-u devnet` shred workflow queries a different program. Decide what `--env devnet` means for this binary, make build_cli_context internally consistent with that choice, and cover the devnet shred-subscription flow in the fork test or CHANGELOG.

crates/solana-cli/src/command/revenue_distribution/fetch/validator_debts.rs
```diff
@@ -58,20 +59,22 @@ struct ValidatorDebtsNodeTableRow {
...
-            solana_connection_options,
             debt_accountant: debt_accountant_key,
+            connection_options,
             dz_env,
         } = self;
 
-        let solana_connection = SolanaConnection::from(solana_connection_options);
+        let solana_connection = crate::command::l1_connection(ctx, connection_options);
+
+        let dz_env_resolved = dz_env.dz_env.or(Some(crate::command::network_env(ctx)));
```
nikw9944: Global `--dz-ledger-url` is ignored here: `dz_env.dz_env.or(Some(network_env(ctx)))` feeds `try_fetch_debt_records_and_distributions`, which builds `DoubleZeroLedgerConnection::from(dz_env)` (env-derived public URL). `ctx.ledger_rpc_url` is never consulted, unlike `fetch distribution` and the shreds verbs. `doublezero-solana --dz-ledger-url http://localhost:8899 revenue-distribution fetch validator-debts` silently queries the public mainnet ledger. Same issue at validator_deposit.rs:148. Plumb `ctx.ledger_rpc_url` through, or document that `--dz-ledger-url` doesn't apply to these verbs.

crates/solana-cli/src/command/shreds/mod.rs
```diff
@@ -22,20 +25,13 @@ use solana_sdk::pubkey::Pubkey;
 
 #[derive(Debug, Args)]
 pub struct ShredsCommand {
```
nikw9944, line 27: The old `--dz-ledger-url` lived on the `shreds` subcommand (`doublezero-solana shreds --dz-ledger-url X list`). It moved to the app level and the global flags are not `global = true`, so the old position now fails to parse — and the removed doc comment said it was "Required for e2e / Docker environments", exactly the scripts most likely to use it. `DZ_LEDGER_URL` env-var users are unaffected. Worth a CHANGELOG line.

crates/solana-cli/src/command/mod.rs
```diff
@@ -20,23 +146,27 @@ pub enum DoubleZeroSolanaCommand {
...
-    pub async fn try_into_execute(self) -> Result<()> {
+    pub async fn execute(self, ctx: &CliContext, out: &mut impl Write) -> Result<()> {
         match self {
-            Self::Passport(passport) => passport.command.try_into_execute().await,
-            Self::RevenueDistribution(revenue_distribution) => {
-                revenue_distribution.command.try_into_execute().await
-            }
-            Self::Shreds(shreds) => shreds.try_into_execute().await,
+            Self::Passport(passport) => passport.command.execute(ctx, out).await,
+            Self::RevenueDistribution(rd) => rd.command.execute(ctx, out).await,
```
nikw9944: `Self::RevenueDistribution(rd)` — `rd` is project-specific shorthand, which CLAUDE.md forbids for variable names anywhere; this was `revenue_distribution` before the refactor. Same spirit applies to the `n` parameter in `map_env`/`network_env` (line 98).

crates/solana-cli/src/command/passport.rs
```diff
@@ -94,71 +94,90 @@ pub struct RequestAdapter {
...
     json: bool,
     json_compact: bool,
 ) -> Result<CliContext> {
-    let env = try_map_env(conn.moniker_env())?;
-    let solana_l1_rpc_url = SolanaConnection::from(conn).url();
+    // Per-verb moniker (back-compat) wins and is validated (passport is not
+    // deployed on Solana devnet); otherwise inherit the global ctx env.
+    let env = match conn.moniker_env() {
+        Some(n) => try_map_env(Some(n))?,
+        None => parent.env,
```
nikw9944: Global `--env devnet` bypasses passport's devnet validation: the per-verb moniker path errors via `try_map_env` ("passport is not available on Solana devnet"), but `merge_ctx` inherits `parent.env` unvalidated, so `doublezero-solana --env devnet passport fetch` proceeds against a cluster where passport isn't deployed while `passport fetch -u devnet` errors. Apply the same validation to the inherited env.

crates/solana-cli/src/main.rs
nikw9944: Help text says "mainnet-beta (default), testnet, localhost", but the value enum accepts `mainnet-beta`, `testnet`, `devnet`, `localnet`. `--env localhost` errors, and `devnet` is undocumented — particularly unfortunate given devnet is the documented route to the testnet shred-subscription program.

crates/solana-cli/src/command/mod.rs
```diff
@@ -4,8 +4,134 @@ mod shreds;
...
+    if opts.solana_url_or_moniker.is_some() {
+        SolanaConnection::from(opts.clone())
+    } else {
+        SolanaConnection::new(ctx.solana_l1_rpc_url.clone())
+    }
+}
+
+/// L1 connection for read verbs: per-verb `--url` (back-compat) wins over the
+/// global ctx L1 URL.
+pub(crate) fn l1_connection(ctx: &CliContext, opts: SolanaConnectionOptions) -> SolanaConnection {
```
nikw9944: `l1_connection` and `shred_subscription_connection` (line 113) are identical implementations (per-verb URL wins, else ctx L1 URL), differing only in by-ref vs by-value. Collapse into one helper. Relatedly, publisher_rewards/configure.rs and prepare_offchain_message.rs build two identical connections from the same options, and passing `Some(shred_connection)` into `build_wallet_with_connection` across the six shreds write verbs quietly reintroduces the redundant second-connection shape PR #396 explicitly collapsed — since both connections now resolve identically, `build_wallet(ctx, opts)` would suffice.

crates/solana-cli/src/main.rs
```diff
@@ -51,7 +97,61 @@ mod tests {
...
 
-    /// Legacy `passport fetch` invocation must still parse unchanged.
+    /// The network environment is resolved solely from `--env`: with `--env
+    /// testnet` and no per-field overrides, the DZ-ledger URL and serviceability
+    /// program ID come from the testnet env config (not from a `--url` moniker),
+    /// and the L1 URL falls back to the env's public Solana URL. Locks in the
+    /// moniker/env decoupling, the one load-bearing behavior change of the
+    /// RFC-20 migration.
+    #[test]
+    fn build_cli_context_resolves_env_defaults_from_env_flag() {
```
nikw9944: The seven new tests here don't start with `test_` as CLAUDE.md requires (`#[test] fn test_<what_it_checks>()`). The preexisting tests in this file share the style, so this is a judgment call — but new code is the usual place to start conforming.

crates/solana-cli/src/command/mod.rs
```diff
@@ -4,8 +4,134 @@ mod shreds;
...
+use doublezero_cli_core::CliContext;
+use doublezero_config::Environment;
+use doublezero_solana_client_tools::{
+    payer::{SolanaPayerOptions, SolanaSignerOptions, Wallet},
+    rpc::{NetworkEnvironment, SolanaConnection, SolanaConnectionOptions},
+};
+
+// ── Shared types & helpers ──────────────────────────────────────────────
+
+/// Per-verb write-transaction options. Connection and keypair come from the
```
nikw9944: Struct doc says "Connection and keypair come from the global `CliContext`; only signer-specific flags remain per-verb", but the struct carries per-verb `connection_options` and `keypair_path` overrides. The doc describes the pre-back-compat design.

crates/solana-cli/src/command/revenue_distribution/relay/distribute_rewards.rs
```diff
@@ -39,10 +39,10 @@ pub struct DistributeRewards {
     schedule: ScheduleOption,
 
     #[command(flatten)]
-    solana_payer_options: SolanaPayerOptions,
+    pub(crate) solana_payer_options: SolanaPayerOptions,
 
     #[command(flatten)]
-    dz_env: DoubleZeroLedgerEnvironmentOverride,
+    pub(crate) dz_env: DoubleZeroLedgerEnvironmentOverride,
```
nikw9944: `dz_env` was widened to `pub(crate)` but is never accessed outside this module — `patch_payer_opts` only touches `solana_payer_options`; the visibility change is dead. Also means relay verbs ignore the global `--env`/`--dz-ledger-url` entirely; presumably acceptable pending #1520, but worth stating in the relay module comment.

crates/solana-cli/src/command/mod.rs
```diff
@@ -20,23 +146,27 @@ pub enum DoubleZeroSolanaCommand {
...
+// ── Shared interactive prompt ───────────────────────────────────────────
+
+pub(crate) fn try_prompt_proceed_confirmation(
+    out: &mut impl Write,
+    prompt_message: &str,
+    abort_message: &str,
+) -> Result<()> {
     loop {
-        println!("⚠️  {prompt_message}. Proceed? [y/N]");
+        writeln!(out, "⚠️  {prompt_message}. Proceed? [y/N]")?;
```
nikw9944, line 136: `try_prompt_proceed_confirmation` writes the prompt to `out` then blocks on `stdin().read_line`. Fine for stdout (line-buffered), but with any buffered writer — the testability target of this refactor — the prompt may never appear before the read blocks. Add `out.flush()?` after the `writeln!`.

crates/solana-cli/src/main.rs
```diff
@@ -28,10 +50,34 @@ async fn main() -> Result<()> {
...
+    let solana_l1_rpc_url = match &app.solana_url {
+        Some(url_or_moniker) => SolanaConnection::from(SolanaConnectionOptions {
+            solana_url_or_moniker: Some(url_or_moniker.clone()),
+        })
+        .url()
+        .to_string(),
+        None => app.env.solana_public_url().to_string(),
+    };
+
+    let mut builder = CliContextBuilder::new()
```
nikw9944: `build_cli_context` never calls `with_client_version`, so `ctx.client_version` is `""`. Nothing reads it in this binary today, but the field's contract says the binary populates it, and the #1520 unified-binary mount will inherit the gap.

nikw9944 (changes requested):
The RFC-20 migration is well executed and verified locally (clippy/fmt clean, 142 lib + 14 binary + 19 client-tools tests pass), but the back-compat story for env-derived values is weaker than the PR description claims. High: (1) trailing per-verb `-u <moniker/url>` no longer drives env-derived values (token mints, oracle key, DZ Ledger URL, serviceability program id) — those come only from global `--env` (default mainnet-beta), so legacy non-mainnet invocations silently resolve mainnet artifacts; this contradicts the PR's back-compat claim, is inconsistent with the passport adapter, and breaks the `-u devnet` workflow PR #396 documented. (2) The devnet↔devnet mapping conflates two taxonomies: `NetworkEnvironment::Devnet` is the Solana devnet cluster while `Environment::Devnet` is the DZ devnet environment living on Solana testnet, so `--env devnet` builds a self-contradictory context and flips device-code resolution from the testnet serviceability program (DZtnuQ…) to the DZ devnet one (GYhQDK…). Relative to PR #396: the DZ-Ledger special-case stays removed, but the redundant second-connection shape #396 collapsed is structurally reintroduced, and #396's documented devnet workflow regresses. Medium: global `--dz-ledger-url` ignored by validator-debt paths; `shreds --dz-ledger-url` flag position removed without changelog; `rd` abbreviation; global `--env devnet` bypasses passport's devnet validation. Several low findings on help text, duplicate helpers, docs, and a missing flush before an interactive prompt.

- Relative to PR #396: the DZ-Ledger special-case removal stands (deleted client-tools helpers stay deleted; no shred traffic routes back to the DZ Ledger), but the redundant second-connection shape #396 collapsed is structurally reintroduced across the shreds write verbs, and #396's documented workflow — reach the testnet shred-subscription program with `-u devnet` — regresses (see the two high findings).
- solana-program-review: not applicable — the PR contains no onchain program code (CLI/offchain plumbing plus two writer-aware output methods in solana-client-tools), so a Solana program review has an empty scope here.

crates/solana-cli/src/command/mod.rs
ben-dz: You're right, and it contradicted the PR's back-compat claim. Fixed by restoring the pre-RFC-20 resolution in every affected verb: per-verb `-u <moniker>` wins, else genesis-hash detection on the connection the verb actually uses (new `resolve_network_env` helper — the passport approach, generalized). Restored wholesale: `serviceability_program_id(env)` (Devnet→testnet id, localnet→"use `--device`" error), `make_dz_connection(url_override, env)`, `DeviceArgs::resolve(env, dz_ledger_url)`, the oracle-key lookup by resolved env, and the `--dz-env` pass-through in validator-debts/validator-deposit. `shreds pay -u devnet --device-code …` again resolves exactly what #396 documented. The global flags now only supply the connection's default URL, so invocations without them behave byte-for-byte as before.

crates/solana-cli/src/command/mod.rs
ben-dz: Agreed. Decision: the global `--env` now takes `doublezero_config::Environment` directly (`mainnet-beta|testnet|devnet|local`) — the DZ taxonomy, matching the `doublezero` CLI these verbs mount into per #1520 — and every unset ctx field derives from `env.config()`. So `--env devnet` builds an internally consistent DZ-devnet context (L1 = Solana testnet, GYhQ… serviceability), and the `map_env`/`network_env` bridges are deleted outright. Per-verb env-derived values come from the moniker/genesis-hash resolution (see the mod.rs:88 thread), so the `-u devnet` route to the testnet shred-subscription program is unchanged. Covered by `test_build_cli_context_devnet_is_doublezero_devnet` + CHANGELOG + `--env` help text.

crates/solana-cli/src/command/revenue_distribution/fetch/validator_debts.rs
ben-dz: Went the other way, deliberately: restored the pre-RFC-20 behavior (pass `--dz-env` through unresolved; these verbs never had a DZ-ledger URL flag), because consuming `ctx.ledger_rpc_url` unconditionally would break the env-derived default whenever the resolved env differs from the ctx default (ctx defaults to mainnet). The global flag's help now documents that it feeds the shreds subcommands (and the passport ctx); plumbing it through revenue-distribution rides with the relay/#1520 migration.

crates/solana-cli/src/command/shreds/mod.rs
ben-dz: Restored at the old position (`shreds --dz-ledger-url <URL> <verb>`) with the old doc comment, threaded to pay/withdraw/list/payments/price as before. The global flag fills the subcommand slot only when it's absent (subcommand-level wins), so both spellings parse. CHANGELOG updated.

crates/solana-cli/src/command/mod.rs
ben-dz: Renamed. (The `n` param went away with `map_env`/`network_env` — both fns are deleted; passport's `try_map_env` now takes `network_environment`.)

crates/solana-cli/src/command/passport.rs
ben-dz: Dissolves with the taxonomy fix (mod.rs:98 thread): `--env devnet` now selects the DZ devnet environment, whose L1 is Solana testnet — where passport is deployed — so there is no unvalidated Solana-devnet path left. The `-u devnet` moniker keeps its explicit error, locked by `test_merge_ctx_devnet_moniker_errors`.

crates/solana-cli/src/main.rs
ben-dz: Fixed — the flag now takes `doublezero_config::Environment`, and the help documents the actual four values (`mainnet-beta`, `testnet`, `devnet`, `local`) plus the devnet-taxonomy note (DZ devnet ≠ Solana devnet cluster; use `-u devnet` for the latter).

crates/solana-cli/src/command/mod.rs
ben-dz: Collapsed into a single `solana_connection`, and the `Some(shred_connection)` shape is gone: all seven shreds write verbs are back to `build_wallet(ctx, opts)` (one connection built from the same per-verb-wins ladder — #396's shape), `build_wallet_with_connection` is deleted, and configure/prepare-offchain-message reuse one connection (configure resolves the mint via `wallet.connection`).

crates/solana-cli/src/main.rs
ben-dz: Renamed — every test in main.rs now carries the `test_` prefix, including the new taxonomy/back-compat ones.

crates/solana-cli/src/command/mod.rs
ben-dz: Rewritten: "Per-verb options for write verbs: the signer-specific flags plus per-verb `--url`/`--keypair` overrides (back-compat) that win over the global `CliContext` values when present."

crates/solana-cli/src/command/revenue_distribution/relay/distribute_rewards.rs
ben-dz: Visibility reverted to private, and the relay module comment now states the limitation explicitly: `patch_payer_opts` merges only the global `--solana-url`/`--keypair`; `--env`/`--dz-ledger-url` are not applied to relay verbs pending the #1520 `Schedulable` migration.

crates/solana-cli/src/command/mod.rs
ben-dz: Added `out.flush()?` after the prompt `writeln!`.

crates/solana-cli/src/main.rs
ben-dz: Populated in `build_cli_context` (`BUILD_VERSION`/`CARGO_PKG_VERSION`) — and it turned out passport's `merge_ctx` was silently dropping it on rebuild, which is now fixed and covered by the new merge_ctx tests.

crates/solana-cli/src/command/passport.rs
ben-dz: Closed it your way: when the per-verb moniker overrides the env, the parent's ledger URL is no longer pinned — the builder derives it from the resolved env, so env/ledger/program-ids stay consistent. Bonus find while writing the test: `merge_ctx` was silently dropping `client_version` (the field-drop class in action) — now carried. Locked in by `test_merge_ctx_moniker_override_rederives_env_fields` and `test_merge_ctx_without_overrides_inherits_parent`.

crates/solana-cli/src/command/shreds/publisher_rewards/prepare_offchain_message.rs
ben-dz: Done — prepare-offchain-message and configure now build a single connection (configure resolves the mint via `wallet.connection`), and the two identical helpers are collapsed into one `solana_connection`.

crates/solana-cli/src/main.rs
ben-dz: Fixed — and `--env` now takes `doublezero_config::Environment` directly, so the help text and the value set come from the same taxonomy the unified binary uses (see the devnet thread on Nik's review for the full decision).

ben-dz:
All review feedback addressed and pushed — see the inline replies. Headline: the back-compat contract is now exact (no behavior change without the new global flags; per-verb moniker/genesis-hash resolution restored everywhere), and the global `--env` uses the DZ environment taxonomy (`--env devnet` = DZ devnet; the Solana devnet cluster stays `-u devnet`). fmt/clippy clean; 145 lib + 16 binary tests pass. @nikw9944 ready for another look.

nikw9944 (dismissed):

juan-malbeclabs (approved):
LGTM

PR #392, opened 2026-06-26
solana-cli: shreds list --all returns only active seats

martinsander00:
## Summary

- `shreds list --all` now returns only seats active in the latest subscription epoch (`active_epoch == max` across the queried seats) and prints that epoch, instead of every seat that has an escrow. `ClientSeat`/escrow accounts persist on-chain after a seat lapses, so the old `--all` count included long-dead seats.

## Testing

- Against mainnet: `shreds list --all` → `Active subscription epoch: 993`, 93 active seats (vs 349 unfiltered before).

### DIFF
--- a/crates/solana-cli/src/command/shreds/list.rs
+++ b/crates/solana-cli/src/command/shreds/list.rs
@@ -38,16 +38,18 @@ pub struct ListCommand {
     #[arg(long)]
     client_ip: Option<Ipv4Addr>,
 
-    /// Show seats regardless of funder, restricted to those active in the latest
-    /// subscription epoch (whose `active_epoch` equals the maximum across the
-    /// queried seats). Lapsed seats, whose accounts persist on-chain, are
-    /// excluded.
+    /// Show seats regardless of funder, restricted to those active in the
+    /// current subscription epoch (whose `active_epoch >= current_epoch`).
+    /// Lapsed seats, whose accounts persist on-chain, are excluded.
     #[arg(long)]
     all: bool,
 
     #[command(flatten)]
     connection_options: SolanaConnectionOptions,
 }
+
+/// A parsed client seat: `(seat_key, device_key, client_ip, tenure)`.
+type ParsedSeat = (Pubkey, Pubkey, Ipv4Addr, u16);
 
 #[derive(Debug, Tabled)]
 struct SeatRow {
@@ -118,15 +120,17 @@ impl ListCommand {
             return Ok(());
         }
 
-        // Parse all seats, recording each seat's active_epoch for the --active
-        // filter below.
+        // Parse all seats. When `--all` is set, also record each seat's
+        // active_epoch for the active-seat filter below.
         let mut active_epoch_by_seat: HashMap<Pubkey, u64> = HashMap::new();
-        let parsed_seats: Vec<_> = accounts
+        let parsed_seats: Vec<ParsedSeat> = accounts
             .iter()
             .filter_map(|(seat_key, account)| {
                 let (device_key, client_ip, tenure, _, active_epoch) =
                     state::parse_client_seat(&account.data)?;
-                active_epoch_by_seat.insert(*seat_key, active_epoch);
+                if self.all {
+                    active_epoch_by_seat.insert(*seat_key, active_epoch);
+                }
                 Some((*seat_key, device_key, client_ip, tenure))
             })
             .collect();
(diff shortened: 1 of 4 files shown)


crates/solana-cli/src/command/shreds/list.rs
```diff
@@ -189,6 +195,23 @@ impl ListCommand {
             (balances, matching_seats)
         };
 
+        let filtered_seats = if self.all {
+            match active_epoch_by_seat.values().copied().max() {
```
nikw9944: "Active" is derived from `max(active_epoch)` over the queried set and matched with `==`, diverging from the rest of the crate, which decides activeness against the authoritative chain epoch with `>=` (`withdraw.rs:126`: `active_epoch >= current_epoch` after `get_epoch_info()`; `pay.rs:312-318`, whose comment explicitly says to use `>=` not `==` because `active_epoch` can run ahead of the reported epoch).

This assumes the queried set always contains an active seat at the true current epoch. It breaks in plausible cases:
- `--all` composes with the `--device`/`--client-ip` memcmp filters (lines 84-101), and `active_epoch_by_seat` is built from the already-filtered result (lines 124-132). `shreds list --all --device-code X` where device X has no active seat → max is a stale past epoch, every lapsed seat matches `== max`, and dead seats are printed as active under a stale "Active subscription epoch" — the very bug this PR fixes.
- Epoch-boundary skew: if one seat is bumped to N+1 mid-settlement while other active seats are at N, max=N+1 drops the still-active N seats.

Recommend `let current_epoch = connection.get_epoch_info().await?.epoch;` and filter `active_epoch >= current_epoch`, printing the real epoch.

crates/solana-cli/src/command/shreds/list.rs
```diff
@@ -115,12 +118,15 @@ impl ListCommand {
             return Ok(());
         }
 
-        // Parse all seats.
+        // Parse all seats, recording each seat's active_epoch for the --active
```
nikw9944: Comment references "the --active filter below", but the flag is `--all`; there is no `--active` flag. Leftover from earlier naming — update to `--all` to avoid confusion.

crates/solana-cli/src/command/shreds/list.rs
```diff
@@ -189,6 +195,23 @@ impl ListCommand {
             (balances, matching_seats)
         };
 
+        let filtered_seats = if self.all {
```
nikw9944, line 202: No test coverage for the new filtering logic, whose `== max` semantics are subtle. `list.rs` has no tests and the filter is buried in a large async fn. Consider extracting a pure helper (e.g. `fn active_seats(seats, current_epoch)`) and unit-testing it, as `pay.rs` does with `is_seat_already_active`/`make_seat_data_ex`. A test would have caught the all-lapsed-subset case above.

crates/solana-cli/src/command/shreds/list.rs
```diff
@@ -115,12 +118,15 @@ impl ListCommand {
             return Ok(());
         }
 
-        // Parse all seats.
+        // Parse all seats, recording each seat's active_epoch for the --active
+        // filter below.
+        let mut active_epoch_by_seat: HashMap<Pubkey, u64> = HashMap::new();
```
nikw9944, line 125: `active_epoch_by_seat` is populated on every `list` invocation but only read inside the `if self.all` branch. Negligible cost, but could be gated behind `self.all` for clarity that it's `--all`-only state.

nikw9944 (changes requested):
The PR restricts `shreds list --all` to seats active in the latest subscription epoch, defining "active" as `active_epoch == max(active_epoch)` over the queried set. Build, clippy, and all tests pass, and the plain `--all` happy path works. However, the activeness check is reimplemented in a way that diverges from the rest of the crate and is fragile.

**High:** The filter infers the current epoch from the data (`max(active_epoch)`) and uses `==`, whereas `withdraw.rs:126` and `pay.rs:318` decide activeness against the authoritative `get_epoch_info().epoch` with `>=`. Because `--all` composes with `--device`/`--client-ip` filters and the max is taken over the already-filtered RPC result, a subset query containing no active seats (e.g. `--all --device-code X` where all of X's seats lapsed) computes a stale max, matches every lapsed seat via `== max`, and reports dead seats as active under a stale "Active subscription epoch" — reintroducing the exact bug this PR fixes. Epoch-boundary settlement skew (which `pay.rs:312` explicitly handles with `>=`) can also drop genuinely-active seats. Recommend filtering `active_epoch >= get_epoch_info().epoch` to match the established pattern.

**Medium:** comments reference a non-existent `--active` flag; no test coverage for the new logic.

**Low:** `active_epoch_by_seat` is built even when `--all` is unset.

crates/solana-cli/src/command/shreds/list.rs
martinsander00: https://github.com/doublezerofoundation/doublezero-offchain/pull/392/changes/c9b0a854d2cef6de6a39fdb7bbe72a8a4698ea13

crates/solana-cli/src/command/shreds/list.rs
martinsander00: https://github.com/doublezerofoundation/doublezero-offchain/pull/392/changes/c9b0a854d2cef6de6a39fdb7bbe72a8a4698ea13

crates/solana-cli/src/command/shreds/list.rs
martinsander00: https://github.com/doublezerofoundation/doublezero-offchain/pull/392/changes/c9b0a854d2cef6de6a39fdb7bbe72a8a4698ea13

crates/solana-cli/src/command/shreds/list.rs
martinsander00: https://github.com/doublezerofoundation/doublezero-offchain/pull/392/changes/c9b0a854d2cef6de6a39fdb7bbe72a8a4698ea13

nikw9944 (dismissed):

nikw9944 (approved):

PR #393, opened 2026-06-30
solana-cli: claim all validator-client-rewards holdings by default

martinsander00:
`shreds validator-client-rewards claim` now drains every outstanding holding by default.

- `--subscription-epoch` is optional. When omitted, all holdings for the client and mint are discovered (current epoch from `getEpochInfo`) and claimed.
- Removes the 16-epoch hard cap: holdings are split into ≤16-per-tx transactions submitted sequentially.
- Explicit epochs whose holding is missing, wrong-mint, or otherwise invalid are warned and skipped instead of failing the whole claim; it only sends nothing when no valid holdings remain.

### DIFF
--- a/crates/solana-cli/src/command/shreds/validator_client_rewards/claim.rs
+++ b/crates/solana-cli/src/command/shreds/validator_client_rewards/claim.rs
@@ -1,4 +1,4 @@
-use anyhow::{Context, Result, anyhow, bail};
+use anyhow::{Context, Result, bail, ensure};
 use clap::Args;
 use doublezero_solana_client_tools::{
     payer::{SolanaPayerOptions, TransactionOutcome, Wallet},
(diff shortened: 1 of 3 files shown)


crates/solana-cli/src/command/shreds/validator_client_rewards/claim.rs
```diff
@@ -89,23 +113,15 @@ impl ClaimCommand {
...
-            .collect();
 
-        // Single getMultipleAccounts call: VCR, ProgramConfig, every holding.
-        let mut all_keys = vec![vcr_key, program_config_key];
-        all_keys.extend(holding_keys.iter().copied());
+        // Single getMultipleAccounts call: VCR, ProgramConfig.
         let accounts = wallet
             .connection
-            .get_multiple_accounts(&all_keys)
+            .get_multiple_accounts(&[vcr_key, program_config_key])
```
karl-dz: There is a try fetch multiple accounts method in client tools that you can use, where you don't have to specify the 100 max const

karl-dz (dismissed):
Looks reasonable to me. Note to self: need to add idiomatic Rust rules to a CLAUDE.md

crates/solana-cli/src/command/shreds/validator_client_rewards/claim.rs
martinsander00: https://github.com/doublezerofoundation/doublezero-offchain/pull/393/changes/e5e982ab43aa8072b65a3ff11ce8f85abedab406

karl-dz (approved):

PR #397, opened 2026-07-07
deps: migrate workspace to Solana 3.0 and ship static musl binaries

ben-dz:
## Summary of Changes
* Workspace `solana-*` crates and `solana-sdk` move to the 3.0 line (`solana-program-test` 3.0.12, `solana-loader-v3-interface` 6, `svm-hash` 0.2, spl interface crates on their v2 lines); doublezero SDK git-deps repin from `client/v0.27.1` to the malbeclabs/doublezero#3830 merge revision `d2824391` (swap for the next v3 `client/*` tag once cut).
* Source adaptations for removed `solana_sdk` re-exports: `solana-sdk-ids`, `solana-compute-budget-interface`, `solana-commitment-config`, `solana-reward-info`, `solana-address-lookup-table-interface`, boxed `ClientErrorKind`, `UiTransactionError` in simulation results, `SysvarSerialize` bounds. All verified semantics-preserving.
* reqwest moves to `rustls-tls` (drops `openssl-sys` from the runtime graph); the CLI and three daemons release as static `x86_64-unknown-linux-musl` binaries so they run on older glibc hosts, mirroring malbeclabs/doublezero#3830. TLS trust roots change from the OS store to the bundled webpki set (changelogged).
* The scheduler NIF stays on the host gnu toolchain (ships inside the Elixir release, not as a cargo artifact).
* Completes the offchain half of the V3 upgrade (malbeclabs/infra#1851); unblocks workspace unification (malbeclabs/infra#1856).
* Fixes https://github.com/malbeclabs/infra/issues/1853

## Testing Verification
* New `rust-cli-static` CI job builds the CLI for musl and asserts `file` reports statically linked; `rust-toolchain.toml` now provisions the musl target and `release/Dockerfile.release` gains `musl-tools`/`cmake` so `scripts/release-rc.sh` snapshot builds keep working.
* `cargo tree` confirms `openssl-sys`/`native-tls` are gone from the runtime dependency graph (openssl remains only via the dev-only `solana-program-test` path).
* Four independent reviews (architecture, security, release-pipeline, SDK) traced every rewritten RPC error-classification path against the Solana 3.0.12 sources and `origin/main`; no critical/high findings, and the one major (RC-snapshot Docker image missing the musl toolchain) is fixed here.
* Fork/e2e coverage relies on the `local-validator` workflow on this branch (already runs against a Solana v3.0.12 runtime).


.github/workflows/ci.yml
```diff
@@ -52,3 +52,32 @@ jobs:
...
+          targets: x86_64-unknown-linux-musl
+      - uses: Swatinem/rust-cache@v2
+      - name: Install musl toolchain
+        run: sudo apt-get update && sudo apt-get install -y musl-tools cmake
+      - name: Build CLI for musl and assert it is statically linked
+        env:
+          CC_x86_64_unknown_linux_musl: musl-gcc
+          CARGO_TARGET_X86_64_UNKNOWN_LINUX_MUSL_LINKER: musl-gcc
+        run: |
+          cargo build --release -p doublezero-solana-cli --target x86_64-unknown-linux-musl
```
juan-malbeclabs: This job only asserts the **CLI** is statically linked, but the goreleaser configs now build four musl artifacts (CLI + contributor-rewards + sentinel + validator-debt). If a daemon later pulls a C-linked dependency, it would produce a dynamically-linked (or unbuildable) release binary that CI never catches — only the release pipeline would. Consider parameterizing this assertion over all four packages, or at least documenting the gap.

.github/workflows/ci.yml
```diff
@@ -52,3 +52,32 @@ jobs:
...
+  # The doublezero-solana CLI ships as a static musl binary so it loads on any
+  # Linux regardless of the host glibc version. Build it for the musl target
+  # and assert the result is fully static.
+  rust-cli-static:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          toolchain: 1.92.0
```
juan-malbeclabs: Toolchain version is duplicated: hardcoded `1.92.0` here and also pinned in `rust-toolchain.toml`. A future bump has to touch both. Note also that `release/Dockerfile.release` still bases on `rust:1.94.0` and relies on `rust-toolchain.toml` overriding to 1.92.0 via rustup inside the container — the mismatch works but is easy to misread.

Cargo.toml
```diff
@@ -61,30 +61,38 @@ metrics-exporter-prometheus = "0.17"
 mockall = "0.13"
 parquet = { version = "57", features = [ "async" ] }
 rayon = "1"
-reqwest = { version = "0.12", features = ["json"] }
+reqwest = { version = "0.12", default-features = false, features = ["json", "rustls-tls"] }
```
juan-malbeclabs: Real behavioral change worth an operational heads-up beyond the changelog: moving from the OS certificate store (openssl/native-tls) to rustls' bundled webpki Mozilla roots means RPC/HTTP endpoints signed by OS-installed private/internal CAs will stop being trusted. This surfaces as a runtime connection failure in specific deployments — worth flagging to whoever operates these daemons before rollout.

Cargo.toml
```diff
@@ -113,42 +121,43 @@ tag = "revenue-distribution/v0.3.7"
...
-# was introduced at client/v0.25.0; the whole SDK family is pinned to a single
-# monorepo revision (client/v0.27.1) so shared types unify across the boundary.
+# render_* output helpers, and the eyre-based error type. The whole SDK family
+# is pinned to a single monorepo revision so shared types unify across the
+# boundary. The pin is the malbeclabs/doublezero#3830 merge commit (Solana 3.0
+# migration); swap for the next v3-containing client/* tag once one is cut.
 [workspace.dependencies.doublezero-cli-core]
 git = "https://github.com/malbeclabs/doublezero"
-tag = "client/v0.27.1"
+rev = "d282439115945aae84128c6baa63cb1d967b1201"
```
juan-malbeclabs, line 132: Pinned to an unreleased merge commit rather than a tag. Acknowledged in the PR body/comment ("swap for the next v3 `client/*` tag once cut"), which is fine to unblock — just a loose end that shouldn't linger: a merge commit is harder to audit than a tag and can be GC'd/rewritten in some workflows.

Cargo.toml
```diff
@@ -61,30 +61,38 @@ metrics-exporter-prometheus = "0.17"
...
+solana-pubkey = { version = "3.0", features = ["borsh", "bytemuck"] }
+solana-reward-info = "3.0"
+solana-rpc-client-types = "3.0"
+solana-sanitize = "3.0"
+# Held to the 3.0 minor: later 3.x releases of the meta-crate remove more
+# deprecated re-exports (sysvar, account, message, ...) that this workspace
+# still consumes.
+solana-sdk = "=3.0"
+solana-sdk-ids = "3.0"
+solana-system-interface = ">=1,<=3"
```
juan-malbeclabs, line 87: Open-ended upper bounds across a major boundary (`>=1,<=3` here, `>=1,<=2` and `>=5,<=6` below) are a minor smell — a fresh `cargo update` could resolve differently than intended. Presumably deliberate to let the git-dep SDK's transitive versions unify without duplicates, and `Cargo.lock` pins the actual resolved versions, so the risk is contained. Flagging, not blocking.

juan-malbeclabs (commented):
Solana 3.0 / static-musl migration. The mechanical import churn and the three non-trivial RPC error-classification rewrites (rpc.rs, transaction.rs, payer.rs) are semantics-preserving, and the changelogs are thorough. A few points inline — the CI static-link coverage and the TLS trust-root change are the two worth resolving before rollout.

.github/workflows/ci.yml
ben-dz: Fixed in d4acdb2: the job now builds all four musl-released packages individually (matching how goreleaser builds each release) and asserts every binary is static. Also verified locally in the release image; all four link static.

.github/workflows/ci.yml
ben-dz: Deduplicated in d4acdb2: both CI jobs now provision via `rustup toolchain install`, which reads rust-toolchain.toml, so the version lives only there. Also added a comment to Dockerfile.release explaining the base-tag vs pinned-toolchain split.

Cargo.toml
ben-dz: Resolved in code in d4acdb2: reqwest now enables rustls-tls-native-roots alongside the webpki bundle, so the OS store is loaded as well and OS-installed private CAs keep working. Sentinel needed a direct reqwest dependency for the feature to unify into its binary. Changelogs updated to match.

Cargo.toml
ben-dz: Agreed, it is a loose end: the pin moves to the next v3-containing client/* tag as soon as one is cut on malbeclabs/doublezero (noted in the Cargo.toml comment).

Cargo.toml
ben-dz: Deliberate: the ranges have to intersect what the doublezero-solana and doublezero SDK git-deps accept so a single version resolves across the boundary, and Cargo.lock pins the outcome. Will tighten once the SDK family settles on v3 tags.

juan-malbeclabs (approved):
All review threads resolved and verified against the code in d4acdb2:

- CI now builds all four musl release packages individually and asserts static linkage on each (scheduler is a prebuilt binary, correctly excluded).
- Toolchain version deduplicated to rust-toolchain.toml; both CI jobs provision via rustup, Dockerfile.release documents the base-tag vs pinned-toolchain split.
- reqwest enables rustls-tls-native-roots alongside the webpki bundle, so OS-installed private CAs stay trusted; sentinel gained a direct reqwest dep for feature unification and all four changelogs document the change.
- Merge-commit pin and open-ended version bounds accepted as documented, non-blocking loose ends to tighten once v3 client/* tags are cut.

LGTM.

PR #398, opened 2026-07-08
solana-cli: shreds pay --operator-key (operator-key split, buyer CLI)

bgm-malbeclabs:
Refs: malbeclabs/infra#1807

## Summary

- Adds `doublezero-solana shreds pay --operator-key <PUBKEY>`, letting a buyer designate a DoubleZero operational identity (the eventual `user.owner` / access-pass `user_payer`) distinct from the escrow's withdraw (finance) authority. This is the buyer-facing half of the operator-key split (PR3 of three: onchain program → oracle → this CLI) for infra#1807, where a customer cannot publish to a private multicast group because their oracle-provisioned shred-seat user is owned by the funding authority rather than their operational identity.
- Carries the operator key on the hand-mirrored `InitializePaymentEscrow` instruction (this repo maintains its own copy of the shred-subscription wire format). Serialization writes `discriminator + 32-byte key`; deserialization treats the key as an **optional trailing argument** (defaults when absent) so `shreds payments` continues to decode transactions that predate the field.
- The flag is optional and defaults to the payer wallet, which reproduces today's behavior exactly (owner = withdraw authority). The operator key is instruction data, so the accounts list is unchanged.

## Testing

- SDK round-trip test for the new tuple form, plus a legacy-wire-form test proving a discriminator-only (pre-field) transaction still decodes (to the default key) — the guarantee `shreds payments` history decoding relies on.
- `cargo test -p doublezero-solana-sdk` (40 pass, incl. the two new tests) and `cargo test -p doublezero-solana-cli` (green); `just clippy` and `just fmt-check` clean; `shreds pay --help` lists `--operator-key`.

## Deploy / rollout

- This CLI must ship **only after** the onchain program change (the shred-subscription `InitializePaymentEscrow` optional-arg upgrade) is deployed. The new CLI unconditionally appends the 32-byte operator key; the currently-deployed program parses instruction data with `try_from_slice`, which rejects trailing bytes, so a new CLI against the un-upgraded program would fail escrow initialization. The upgraded program accepts both old and new CLIs. This is the one hard ordering constraint for the coordinated rollout.

### DIFF
--- a/crates/solana-cli/src/command/shreds/pay.rs
+++ b/crates/solana-cli/src/command/shreds/pay.rs
@@ -117,9 +117,13 @@ fn user_owner_is_acceptable(
     user_owner: Option<Pubkey>,
     oracle_key: Option<Pubkey>,
     wallet_key: Pubkey,
+    operator_key: Pubkey,
 ) -> bool {
     let is_shred_oracle_user = oracle_key.zip(user_owner).is_some_and(|(o, u)| o == u);
-    let is_self_owned = user_owner == Some(wallet_key);
+    // Under the operator-key split the oracle provisions the user owned by the
+    // operator key, which may be distinct from the wallet (withdraw authority);
+    // a re-subscribe/top-up must accept either as self-owned.
+    let is_self_owned = user_owner == Some(wallet_key) || user_owner == Some(operator_key);
     is_shred_oracle_user || is_self_owned
 }
 
@@ -232,7 +236,8 @@ impl PayCommand {
                     None
                 };
 
-                if !user_owner_is_acceptable(user_owner, oracle_key, wallet.pubkey()) {
+                if !user_owner_is_acceptable(user_owner, oracle_key, wallet.pubkey(), operator_key)
+                {
                     bail!(
                         "Client IP {} already has a multicast user on serviceability \
                          owned by neither the shred oracle nor your wallet. This IP \
@@ -1001,7 +1006,12 @@ mod tests {
     fn user_acceptable_when_owned_by_shred_oracle() {
         let oracle = Pubkey::new_unique();
         let wallet = Pubkey::new_unique();
-        assert!(user_owner_is_acceptable(Some(oracle), Some(oracle), wallet,));
+        assert!(user_owner_is_acceptable(
+            Some(oracle),
+            Some(oracle),
+            wallet,
+            wallet,
+        ));
     }
 
     #[test]
@@ -1009,14 +1019,35 @@ mod tests {
         // New behavior: validator-owned (self-owned) Users are benign.
         let oracle = Pubkey::new_unique();
         let wallet = Pubkey::new_unique();
-        assert!(user_owner_is_acceptable(Some(wallet), Some(oracle), wallet,));
+        assert!(user_owner_is_acceptable(
+            Some(wallet),
+            Some(oracle),
+            wallet,
+            wallet,
+        ));
+    }
+
+    #[test]
+    fn user_acceptable_when_owned_by_operator_key() {
+        // Operator-key split: the oracle provisions the user owned by the
+        // operator key, distinct from the wallet (withdraw authority). A
+        // re-subscribe/top-up must treat that as self-owned.
+        let oracle = Pubkey::new_unique();
+        let wallet = Pubkey::new_unique();
+        let operator = Pubkey::new_unique();
+        assert!(user_owner_is_acceptable(
+            Some(operator),
+            Some(oracle),
+            wallet,
+            operator,
+        ));
     }
 
     #[test]
     fn user_acceptable_self_owned_even_when_oracle_key_unknown() {
         // If we can't resolve the oracle pubkey, self-ownership still passes.
         let wallet = Pubkey::new_unique();
-        assert!(user_owner_is_acceptable(Some(wallet), None, wallet));
+        assert!(user_owner_is_acceptable(Some(wallet), None, wallet, wallet));
     }
 
     #[test]
@@ -1028,6 +1059,7 @@ mod tests {
             Some(third_party),
             Some(oracle),
             wallet,
+            wallet,
         ));
     }
 
@@ -1036,13 +1068,23 @@ mod tests {
         // Malformed account data → user_owner is None → bail (safe default).
         let oracle = Pubkey::new_unique();
         let wallet = Pubkey::new_unique();
-        assert!(!user_owner_is_acceptable(None, Some(oracle), wallet));
+        assert!(!user_owner_is_acceptable(
+            None,
+            Some(oracle),
+            wallet,
+            wallet
+        ));
     }
 
     #[test]
     fn user_rejected_when_third_party_and_oracle_unknown() {
         let wallet = Pubkey::new_unique();
         let third_party = Pubkey::new_unique();
-        assert!(!user_owner_is_acceptable(Some(third_party), None, wallet));
+        assert!(!user_owner_is_acceptable(
+            Some(third_party),
+            None,
+            wallet,
+            wallet,
+        ));
     }
 }
--- a/crates/solana-cli/src/command/shreds/pay.rs
+++ b/crates/solana-cli/src/command/shreds/pay.rs
@@ -171,6 +171,11 @@ pub struct PayCommand {
     /// Serviceability program ID for the multicast user guard (auto-detected; override for e2e)
     #[arg(long, hide = true)]
     serviceability_program_id: Option<Pubkey>,
+    /// Operator key that becomes the DoubleZero operational identity
+    /// (user.owner / access-pass user_payer). Defaults to the payer wallet,
+    /// reproducing today's behavior (owner = the withdraw authority).
+    #[arg(long)]
+    operator_key: Option<Pubkey>,
 
     #[command(flatten)]
     write_opts: crate::command::WriteVerbOptions,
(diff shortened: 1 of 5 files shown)


crates/solana-cli/src/command/shreds/pay.rs
```diff
@@ -415,7 +421,7 @@ impl PayCommand {
             let escrow_ix = try_build_instruction(
                 &ID,
                 InitializePaymentEscrowAccounts::new(&client_seat_key, &wallet_key),
-                &ShredSubscriptionInstructionData::InitializePaymentEscrow,
+                &ShredSubscriptionInstructionData::InitializePaymentEscrow(operator_key),
```
nikw9944, line 461: **H1:** `--operator-key` is silently ignored when the escrow already exists. The operator key is carried only by `InitializePaymentEscrow`, built inside `if !escrow_exists` (line 420), and the SDK has no instruction to set/change it on an existing escrow. On a top-up run with `--operator-key X`, the transaction succeeds with no warning and no output of the effective key — the user's operational identity (`user.owner` / access-pass `user_payer`) is silently not applied, reproducing the multicast-publish failure infra#1807 targets. Fix: `bail!` (or prominently warn) when `self.operator_key.is_some() && escrow_exists`, and print the effective operator key when the escrow is initialized.

crates/solana-sdk/src/shred_subscription/instruction/mod.rs
```diff
@@ -187,7 +191,11 @@ impl BorshDeserialize for ShredSubscriptionInstructionData {
                 let client_ip = u32::deserialize_reader(reader)?;
                 Ok(Self::InitializeClientSeat { client_ip })
             }
-            Self::INITIALIZE_PAYMENT_ESCROW => Ok(Self::InitializePaymentEscrow),
+            Self::INITIALIZE_PAYMENT_ESCROW => {
+                // Optional trailing argument: absent (legacy tx) -> default key.
+                let operator_key = Pubkey::deserialize_reader(reader).unwrap_or_default();
```
nikw9944: **M1:** `unwrap_or_default()` swallows every error, not just clean EOF, and the truncated case is reachable: std's `<&[u8] as Read>::read_exact` consumes all remaining bytes before returning its EOF error, so borsh's `try_from_slice` "Not all bytes read" guard passes vacuously (verified empirically against locked borsh 1.5.7 / solana-pubkey 2.4.0). `discriminator + 1..31 garbage bytes` decodes to `Ok(InitializePaymentEscrow(Pubkey::default()))` instead of `Err`; only ≥33 trailing bytes error. The onchain program requires the key be absent or exactly 32 bytes, so this mirror decodes byte strings the chain rejects, and malformed instructions land in payments.rs's benign match arm instead of `Err(_)`. Fix: read the remaining bytes and require exactly 0 (legacy → default) or 32; anything else is `io::ErrorKind::InvalidData`.

crates/solana-sdk/src/shred_subscription/instruction/mod.rs
```diff
@@ -259,6 +267,27 @@ mod tests {
         assert_eq!(*ix, parsed);
     }
 
+    #[test]
+    fn round_trip_initialize_payment_escrow() {
```
nikw9944: **M2:** No frozen-bytes test pins the new wire form (discriminator immediately followed by the 32 pubkey bytes). The module has this pattern for exactly this reason (`frozen_bytes_initialize_claim_holding`, `frozen_bytes_claim_validator_client_rewards_one_entry`); a round-trip test cannot catch a symmetric encode/decode bug, and this layout is the contract with the deployed program. Add coverage asserting `borsh::to_vec(&InitializePaymentEscrow(key)) == discriminator bytes ++ key.to_bytes()`.

nikw9944: **M3:** Repo-root CLAUDE.md: "**Every test function name starts with `test_`.** Write `#[test] fn test_<what_it_checks>()`, not `#[test] fn <what_it_checks>()`." The new `round_trip_initialize_payment_escrow` (line 271) and `initialize_payment_escrow_decodes_legacy_wire_form` (line 281) break it. The pre-existing tests in this module also break it, so renaming only the new ones trades file-local consistency for rule compliance — resolve deliberately.

crates/solana-cli/src/command/shreds/pay.rs
```diff
@@ -185,6 +190,7 @@ impl PayCommand {
             .into_shred_subscription_connection();
         let wallet = Wallet::try_new(self.solana_payer_options, Some(dz_connection))?;
         let wallet_key = wallet.pubkey();
+        let operator_key = self.operator_key.unwrap_or(wallet_key);
```
nikw9944, line 204: **L1:** `operator_key` is bound here but used exactly once ~230 lines later (line 424). Inline `self.operator_key.unwrap_or(wallet_key)` at the call site, matching how `source_token_account` is resolved next to its use.

nikw9944 (changes requested):
Small, well-tested change: the SDK's hand-mirrored `InitializePaymentEscrow` gains a `Pubkey` payload and `shreds pay` gains `--operator-key` (defaulting to the payer wallet). Both crates' tests pass; clippy and rustfmt are clean; the deploy-ordering constraint is documented. Requesting changes for one high-severity gap: `--operator-key` is silently ignored on the escrow top-up path, with no warning and no way to apply the key afterward — the user's operational identity is silently not set, reproducing the exact failure infra#1807 targets. Also confirmed (empirically, against locked borsh 1.5.7 / solana-pubkey 2.4.0) that the new optional-trailing-arg decode accepts truncated 1–31-byte payloads as `Pubkey::default()` because `<&[u8]>::read_exact` consumes remaining bytes before erroring, defeating borsh's leftover-bytes check; the onchain program is strict here, so the mirror diverges from the chain. Plus a missing frozen-bytes test for the new wire form and two convention nits.

- The hard rollout ordering (upgraded program before this CLI, else escrow init fails with an opaque onchain error) is acknowledged in the PR description; no in-code gate exists, which is acceptable given the stated coordinated rollout.

crates/solana-cli/src/command/shreds/pay.rs
bgm-malbeclabs: Kept it as a binding — `operator_key` has a second use site at pay.rs:238, in the `user_owner_is_acceptable` multicast-user collision check, in addition to the escrow init. Inlining `self.operator_key.unwrap_or(wallet_key)` would duplicate the fallback across the two, so the `let` binding stays. Addressed H1/M1/M2/M3 in b2bd2fe.

crates/solana-cli/src/command/shreds/pay.rs
```diff
@@ -411,13 +422,29 @@ impl PayCommand {
             compute_unit_limit += 50_000 + Wallet::compute_units_for_bump_seed(seat_bump);
         }
 
+        // The operator key is only carried by InitializePaymentEscrow and is
+        // immutable once the escrow exists — there is no set-operator-key
+        // instruction. Silently ignoring it on a top-up run would leave the
+        // user's operational identity unset, reproducing the multicast-publish
+        // failure infra#1807 targets. Refuse rather than mislead.
+        if escrow_exists && self.operator_key.is_some() {
```
nikw9944: **H1-NEW (adversarially verified, CONFIRMED):** This bail combines with 95c5ca7 to block every top-up of an operator-keyed seat. The oracle provisions the User PDA owned by operator key X; the escrow PDA is keyed by `(client_seat_key, wallet_key)`, so a same-wallet top-up always has `escrow_exists = true`. With `--operator-key X`, the guard passes (line 126 accepts `user_owner == operator_key`) but this check bails even though the user is re-stating the same key. Without the flag, `operator_key` defaults to `wallet_key` (line 198) and the guard at line 239 rejects `user_owner = X` as "owned by neither the shred oracle nor your wallet". Both paths bail — 95c5ca7's own commit message says the guard change exists to let this flow through. `PaymentEscrow` state stores no operator key (state.rs:409), so the CLI can't cross-check a stored value. Suggested CLI-only fix: when `escrow_exists` and the flag is passed, allow it if it matches the User PDA owner already fetched by the guard (bail only on a genuine change attempt); and/or relax the without-flag guard when the wallet's own escrow for this seat exists.

nikw9944 (changes requested):
Resolution check at b2bd2fe: four of five prior findings resolved (M1 strict 0-or-32-byte decode + truncation test, M2 frozen-bytes test, M3 test_ prefixes, H1 bail + effective-key output) and L1 was validly declined (operator_key now has two use sites). However, the two new commits interact to create a new high-severity regression: top-ups of a seat provisioned with a distinct operator key are blocked on every path. With `--operator-key X` the guard passes but the new `escrow_exists` bail fires; without the flag, operator_key defaults to the wallet and the multicast-user guard rejects the operator-owned User PDA. Both paths bail, so exactly the infra#1807 customers this PR serves cannot re-fund once provisioned. Tests (146+16+42) and clippy are clean at the new head.

- Prior findings resolved: M1 (decode now reads to end and requires exactly 0 or 32 trailing bytes, with a truncated-key rejection test), M2 (frozen-bytes test pins discriminator ++ 32 key bytes), M3 (new tests carry the test_ prefix), H1 mechanics (bail on top-up with flag + effective operator key printed on init) — though the bail's interaction is the new H1-NEW finding.
- L1 declined with valid rationale: after 95c5ca7, operator_key has two use sites (multicast-user guard and escrow init), so the early binding is justified.
- Verified at new head: cargo test for both crates passes (146+16+42) and clippy is clean; the branch was rebased onto current main (RFC-20 verb refactor etc.) with the PR commits applied on top.

nikw9944 (approved):
All posted findings have been resolved

PR #401, opened 2026-07-20
Revert shreds pay --operator-key (operator-key split buyer CLI)

martinsander00:
Refs: malbeclabs/infra#1807

Backs out the buyer-CLI half of the operator-key split, which is being replaced by a simpler allowlist-based approach (see the doublezero-shreds revert). Coordinated with that PR.

## Summary

- Reverts #398: the `--operator-key` flag on `shreds pay` and the operator-key argument in the `solana-sdk` `InitializePaymentEscrow` builder.

## Deploy / rollout

- Ship together with the doublezero-shreds revert: the reverted `shred-subscription` program rejects the trailing operator-key bytes this CLI appended, so a non-reverted CLI against the reverted program would fail escrow creation.

### DIFF
--- a/crates/solana-cli/src/command/shreds/pay.rs
+++ b/crates/solana-cli/src/command/shreds/pay.rs
@@ -117,19 +117,9 @@ fn user_owner_is_acceptable(
     user_owner: Option<Pubkey>,
     oracle_key: Option<Pubkey>,
     wallet_key: Pubkey,
-    operator_key: Pubkey,
-    wallet_owns_escrow: bool,
 ) -> bool {
     let is_shred_oracle_user = oracle_key.zip(user_owner).is_some_and(|(o, u)| o == u);
-    // Under the operator-key split the oracle provisions the user owned by the
-    // operator key, which may be distinct from the wallet (withdraw authority);
-    // a re-subscribe/top-up must accept either as self-owned. When this wallet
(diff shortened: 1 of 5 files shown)


crates/solana-cli/src/command/shreds/pay.rs
```diff
@@ -283,6 +239,23 @@ impl PayCommand {
...
+        // Check which accounts already exist on-chain.
+        let accounts = wallet
+            .connection
+            .get_multiple_accounts(&[client_seat_key, escrow_key, program_config_key])
+            .await?;
+        let seat_exists = accounts[0].is_some();
+        let escrow_exists = accounts[1].is_some();
+        let prorated_service_enabled = accounts[2]
+            .as_ref()
+            .is_some_and(|a| state::is_prorated_service_enabled(&a.data));
```
karl-dz: Might be more idiomatic to use `into_iter()` with `next().is_some()` for each of these. But this is fine

karl-dz (dismissed):

crates/solana-cli/src/command/shreds/pay.rs
martinsander00: https://github.com/doublezerofoundation/doublezero-offchain/pull/401/changes/ec0a8c6731457c240a3a33a26839fd1f8db19499

karl-dz (approved):

PR #404, opened 2026-07-22
contributor-rewards: warn and skip undecodable serviceability accounts

### LINKED ISSUE #403: contributor-rewards: warn+skip undecodable serviceability accounts
# contributor-rewards: warn+skip undecodable serviceability accounts

Repo: `doublezerofoundation/doublezero-offchain`. Single small PR.

## Problem

`serviceability::fetch` (`crates/contributor-rewards/src/ingestor/serviceability.rs`) decodes each fetched account with `X::try_from(&account_data[..])?` inside the per-account loop. The `?` propagates, so **one undecodable account fails the entire epoch snapshot** — every epoch, forever, since the scheduler retries the same epoch.

This bit hard in production on 2026-07-22: testnet AccessPass `5x9DTsWCrAQtYKH4W7trEjoaCRHnzywsZp2oFX3sg5Zg` was written under the new `EdgeSeat(Vec<FeedSeat>)` schema (malbeclabs/doublezero#3954) while the deployed contributor-rewards pinned a pre-FeedSeat SDK; the misparse triggered a 52 GB allocation abort and `do-tn-rewards1` crash-looped on epoch 267. The SDK pin is fixed by doublezero-offchain#402 and SDK-side allocation hardening is a separate PR in malbeclabs/doublezero. **This PR is the ingest-policy layer: a decode failure on one account must degrade gracefully, not kill the epoch.**

`telemetry::fetch` (`crates/contributor-rewards/src/ingestor/telemetry.rs`) already does this correctly — per-account failures are logged and counted, processing continues. Bring serviceability to parity.

## Deliverables

1. In `serviceability::fetch`: per-account decode failure → `warn!` (include pubkey, account type, `account_data.len()`, and the error) → skip that account → continue. No `?` out of the per-account loop.
2. Count decode failures distinctly from the existing per-type fetch failures, and include the decode-failure count in the completion `info!` line (it already logs `Errors={}` for per-type fetch errors — keep both visible; don't conflate them).
3. Tests (required — this is the "test that would have caught it"): a batch containing one valid account and one corrupt/truncated account of the same type must yield the valid account in `DZServiceabilityData`, count one decode error, and return `Ok`. Structure however fits the existing test layout — if the per-account decode dispatch needs to be separable from the RPC fetch to be testable, extracting it is acceptable (it's the unit under test). Use real serialized bytes for the valid account (e.g. `borsh::to_vec` of a constructed state struct, as the SDK's own tests do), and truncated bytes for the corrupt one.

## Constraints

- Behavior note for the PR description: skipping an account silently alters rewards input, which is why the skip must be a loud `warn!` with the pubkey and a counter in the completion log — operators must be able to see it. Do not add config to toggle this behavior; skip-and-warn is the policy.
- Do not touch the fetch/retry/RPC layers, the SDK pin, or other ingestors.
- Follow repo `CLAUDE.md` conventions strictly (no inferable type annotations, `anyhow::Context` over `anyhow!`, `test_` prefix, no new crates, no single-caller helpers except units under test).
- PR title style: `contributor-rewards: warn and skip undecodable serviceability accounts` (adjust wording, keep the `component:` form). Reference doublezero-offchain#402 and the incident context above in the description.
- Keep the diff focused (this is a policy change plus tests, not a refactor), and deliver it as ONE PR — never split the work into multiple PRs.

ben-dz:
## Summary of Changes
* `serviceability::fetch` no longer aborts the whole fetch mid-loop when one account fails to decode. The per-type account loop is extracted into `decode_accounts`, which warns (pubkey, account type, byte length, error), counts, and skips each account whose `try_from` returns `Err`, then returns the processed count and the failed pubkeys.
* **Policy split (reward-bearing vs. reward-neutral):** an AccessPass decode failure is warned and skipped — AccessPass does not affect rewards. For every other type (Location, Exchange, Device, Link, User, MulticastGroup, Contributor), `fetch` completes the full sweep so every bad account gets a warn line, then returns `Err` naming the affected types, counts, and pubkeys. A partial snapshot on reward-bearing types would feed Shapley a silently shrunk graph (builders drop devices/links on missing cross-references), post a skewed merkle, and `rewards_exist_for_epoch` would freeze that epoch permanently. A loud failed epoch is recoverable; a frozen skewed merkle is not.
* Each decode failure increments a Prometheus counter `doublezero_contributor_rewards_serviceability_decode_errors` (labeled by `account_type`), so even the skipped-AccessPass path can back an alert. The completion `info!` line still reports `Errors=` (per-type fetch failures) and `DecodeErrors=` (total decode failures) distinctly.
* Fixes #403

## Scope / what this does and does not fix
This is the ingest-policy layer only. It handles decode failures that surface as an `Err` from the SDK's `try_from`. It does **not** and cannot catch the 52 GB allocation abort from the 2026-07-22 incident — an allocation failure aborts the process before `try_from` can return, so it is not a catchable `Err`. That class is addressed by the SDK pin bump in #402 (merged) and by allocation hardening in malbeclabs/doublezero#4074. Given the pinned SDK's lenient `try_from` (per-field `unwrap_or_default`, `Err` only on a leading-discriminant mismatch) plus the offset-0 memcmp filter, most fetched accounts decode `Ok`; `Device::try_from` is the one processed type that can still `Err` mid-parse. The value here is a defined, loud policy for the decode failures that do surface, and parity of intent with `telemetry::fetch`.

## Testing Verification
* One mixed-batch unit test on the extracted `decode_accounts`: a valid `Location` (built inline via `borsh::to_vec`) plus one undecodable account. Asserts the valid account landed in the locations map, exactly one decode failure is reported with the garbage pubkey, and the garbage account is not inserted. The test exercises the real skip-and-count loop (not a loop re-implemented in the test body), so a regression to `?` mid-loop would fail it.
* `cargo test -p doublezero-contributor-rewards --lib` (101 passed), `cargo clippy --all-targets` clean, `cargo fmt` applied.

## Review history
Earlier self-review rounds (design + security sub-agents) are recorded in the repo artifacts. This revision supersedes the initial blanket skip-and-warn policy with the reward-bearing/reward-neutral split above, per the plan owner's decision, and folds the per-type loop into the tested unit.

### DIFF
--- a/crates/contributor-rewards/CHANGELOG.md
+++ b/crates/contributor-rewards/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - TLS for HTTP clients moves from openssl to rustls; trust roots are the bundled webpki Mozilla set plus the host OS certificate store, so OS-installed private CAs remain trusted (malbeclabs/infra#1853)
 - refactor(contributor-rewards): use the shared `Wallet` memo helpers from `solana-client-tools` in place of the local `RELAY_MEMO_CU` constant
 - fix(contributor-rewards): backfill `User.feed_pk` (zero pubkey) in the snapshot compat migration so snapshots captured before doublezero#4030 still deserialize (malbeclabs/infra#1853)
+- fix(contributor-rewards): warn and skip an undecodable serviceability account instead of failing the whole epoch snapshot; decode failures are counted separately and surfaced as `DecodeErrors=` in the completion log ([#403](https://github.com/doublezerofoundation/doublezero-offchain/pull/403))
 
 ## [0.6.1](https://github.com/doublezerofoundation/doublezero-offchain/releases/tag/contributor-rewards%2Fv0.6.1) - 2026-06-13
 
--- a/crates/contributor-rewards/CHANGELOG.md
+++ b/crates/contributor-rewards/CHANGELOG.md
@@ -12,7 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - TLS for HTTP clients moves from openssl to rustls; trust roots are the bundled webpki Mozilla set plus the host OS certificate store, so OS-installed private CAs remain trusted (malbeclabs/infra#1853)
 - refactor(contributor-rewards): use the shared `Wallet` memo helpers from `solana-client-tools` in place of the local `RELAY_MEMO_CU` constant
 - fix(contributor-rewards): backfill `User.feed_pk` (zero pubkey) in the snapshot compat migration so snapshots captured before doublezero#4030 still deserialize (malbeclabs/infra#1853)
-- fix(contributor-rewards): warn and skip an undecodable serviceability account instead of failing the whole epoch snapshot; decode failures are counted separately and surfaced as `DecodeErrors=` in the completion log ([#403](https://github.com/doublezerofoundation/doublezero-offchain/pull/403))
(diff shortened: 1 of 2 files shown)


martinsander00 (approved):
i like this

PR #405, opened 2026-07-26
shreds: price the pay preflight from last_settled_epoch

ben-dz:
## Problem

`shreds pay`'s price floor read the **newest** metro/device ring entry, but `request_instant_seat_allocation` charges from the entries at the execution controller's **`last_settled_epoch`** — during `OpenForRequests` that is at least one epoch behind. While prices are static the two agree and nothing shows. The epoch a metro reprices, the preflight passes an underfunded amount straight through to an opaque `invalid account data for instruction`, which is precisely what the guard exists to prevent.

Confirmed incident: testnet `lax-dz001` (`CT8mP6RUoRcAB67HjKV9am7SBTCpxaJEwfQrSjVLdZfD`) was repriced 43 -> 10; after the epoch rolled on 2026-07-24, `shreds price` quoted 10 while the program charged 43 prorated, failing three `qa.testnet` runs (`Payment escrow must have at least 41034841 / 39996869 / 34028528 USDC`). Mainnet has the same exposure — 15 devices across 7 metros dropped from 30/60 to 10.

## Changes

- **`solana-sdk` `shred_subscription/state.rs`** — epoch lookup into the DeviceHistory/MetroHistory ring buffers (`parse_metro_history_price_at_epoch`, `parse_device_history_premium_at_epoch`), mirroring the onchain `RingBuffer::find`: walk backwards from `current_index`, bounded by `total_count`, so `epoch == 0` cannot match a zero-initialized slot. `seat_usdc_price_dollars` combines the base price with the signed premium the way `DeviceSubscription::usdc_price_dollars` does. Existing `parse_metro_history` / `parse_device_history` are untouched — the contributor-rewards ingestor and `shreds list` legitimately want the newest entry.
- **`shreds pay`** — appends the execution controller to the existing batch fetch (no extra round trip) and computes the floor from `last_settled_epoch`. A missing ring entry now refuses to submit with a clear diagnostic instead of falling back to the newest entry; the program errors in that case too. The rejection message names both prices and both epochs:
  ```
  Amount (10.000000 USDC) is below the seat price charged for this epoch (43.000000 USDC).
    The seat you are buying covers the remainder of epoch 946, priced at 43 USDC.
    The price changes to 10 USDC in epoch 947.
  ```
  The per-seat price override keeps precedence, mirroring `checked_override_usdc_price_dollars().unwrap_or(default)`.
- **`shreds price`** — new `Instant Price (USDC)` column / `instant_allocation_price` JSON field: what `pay` charges right now. `epoch_price` keeps its existing meaning (what a recurring subscriber pays next epoch). One extra key in the existing `try_join!` batch.

The floor stays **unprorated**, as before. With `prorated_service_enabled` the program charges less, so the floor is stricter than strictly necessary — but it is always >= the program's requirement, so it can never cause the revert this fixes, and any surplus lands in the escrow. Proration would need `get_first_slot_in_epoch(current_subscription_epoch)`, which can be fast-forwarded past `clock.epoch + 1`; getting that subtly wrong would make the floor too low and reintroduce the revert.

## Verification

- Unit tests for the ring search: exact hit, wraparound past index 0, the `total_count` bound, `epoch == 0` against both a zero-initialized and a written buffer, miss, truncated buffer.
- Unit tests for the floor: the incident itself (settled 43, current 10, `--amount 10` rejected with the message above), signed premium applied at the settled epoch, override precedence, and no divergence line when the price is stable.
- Live against DZ testnet, which still shows the divergence:
  ```
  | lax-dz001 | xlax | Los Angeles | activated | 0 | 64 | 10 | 0 | 10 | 43 |
  ```
  `epoch_price` 10, `instant_allocation_price` 43. Mainnet shows the same for the Mumbai devices (10 vs 30). Both converge once the epoch rolls.

## Follow-up (not in this PR)

`malbeclabs/doublezero` `e2e/qa_shred_settlement_test.go` reads `epoch_price` from `shreds price --json` and pays it, so it will keep failing until it reads `instant_allocation_price` instead — but it now fails with the message above rather than a raw simulation dump.


crates/solana-cli/src/command/shreds/pay.rs
```diff
@@ -364,29 +441,56 @@ impl PayCommand {
...
+        )
+        .with_context(|| {
+            format!(
+                "Device history {device_history_key} has no subscription for the last settled \
+                 epoch {last_settled_epoch}, so the instant seat allocation would be rejected \
+                 onchain"
+            )
+        })?;
+
+        if let Some(message) = seat_price_shortfall_message(&SeatPriceInput {
```
nikw9944, line 511: This floor (and the two ring-presence hard-errors above at lines 461-481) runs unconditionally, but `RequestInstantSeatAllocation` is only added when `!seat_already_active` (line 551). For an active seat the transaction is a pure `FundPaymentEscrowUsdc` top-up with no onchain minimum, so after a price decrease (43 -> 10, the incident scenario) topping up at the correct next-epoch price of 10 is now rejected demanding 43 — pre-PR the newest-entry floor allowed it, and settlement charges the newest price. The message ("The seat you are buying covers the remainder of epoch ...") and the ring-entry diagnostics ("the instant seat allocation would be rejected onchain") are also wrong for that path. `seat_already_active` is already computed at line 382: apply the settled-epoch floor and ring-presence requirements only when `!seat_already_active`, keeping the newest-entry floor (or no floor) for top-ups.

crates/solana-cli/src/command/shreds/price.rs
```diff
@@ -170,17 +185,30 @@ impl PriceCommand {
...
             connection.try_fetch_multiple_accounts(&all_history_keys),
             dz_connection.try_fetch_multiple_accounts(&exchange_keys),
         )?;
 
-        let (dh_data, mh_data) = history_accounts.split_at(dh_keys.len());
+        let (dh_data, rest) = history_accounts.split_at(dh_keys.len());
+        let (mh_data, execution_controller_data) = rest.split_at(mh_keys.len());
+
+        // The epoch an instant seat allocation is priced from.
+        let last_settled_epoch = execution_controller_data
```
nikw9944: `shreds price` is a read-only listing that previously never touched the execution controller; now a missing/unparseable controller (fresh localnet/e2e deployment, layout shift) aborts the whole command. `instant_allocation_price` is already `Option<u16>` rendering as `-` to tolerate missing per-device data — treat `last_settled_epoch` as optional too, skip the settled-price maps and the explanatory line when absent, and keep the table. Hard-erroring is right for `pay`, not needed here.

crates/solana-cli/src/command/shreds/pay.rs
```diff
@@ -244,11 +307,18 @@ impl PayCommand {
         let (escrow_key, escrow_bump) =
             state::find_payment_escrow_address(&client_seat_key, &wallet_key);
         let (program_config_key, _) = state::find_program_config_address();
+        let (execution_controller_key, _) = state::find_execution_controller_address();
 
-        // Check which accounts already exist on-chain.
+        // Check which accounts already exist on-chain, and read the epoch the
```
nikw9944: This rewritten comment (and line 445) says "on-chain"; repo CLAUDE.md: "Always write 'onchain', never 'on-chain'." The PR's other new prose gets this right.

nikw9944 (commented):
The settled-epoch price floor needs to be scoped to the instant-allocation path: as written it also gates pure escrow top-ups (seat already active), where no `RequestInstantSeatAllocation` is submitted, so after a price decrease — the exact incident scenario — an active seat holder topping up at the correct next-epoch price is rejected demanding the old settled price, with a message describing a seat purchase they are not making. Also: the `changelog` CI check is failing (no entries for the touched crates), and `shreds price` now aborts entirely when the execution controller is missing instead of degrading the already-optional Instant Price column.

- The `changelog` CI check is failing: the PR touches `crates/solana-cli` and `crates/solana-sdk` but updates neither crate's CHANGELOG.md. This is a user-visible CLI behavior change (new `Instant Price` column, new rejection semantics) of the kind prior entries document (e.g. the #395 `shreds withdraw` entry). Add `[Unreleased]` entries or the `skip-changelog` label.

crates/solana-cli/src/command/shreds/pay.rs
```diff
@@ -102,6 +102,69 @@ fn epoch_warning_prompt(input: &EpochWarningInput) -> Option<String> {
     ))
 }
 
+/// Inputs for the instant-allocation price floor, separated from I/O for
```
nikw9944: `///` rustdoc on a new private struct — repo CLAUDE.md reserves `///` for the published API and forbids it on internal structs. Switch to `//`; the content is worth keeping.

ben-dz:
Thanks — the top-up regression was real, all three addressed in bf43d77.

**Scoping the floor to the instant-allocation path.** You're right: an already-active seat submits no `RequestInstantSeatAllocation`, so the pay is a pure escrow top-up and the next settlement charges the *newest* ring entry. The floor now branches on `seat_already_active`: the purchase path prices from `last_settled_epoch`, the top-up path keeps the newest entry (pre-PR behavior for that path) and gets its own message that talks about funding the escrow rather than buying a seat. The hard failure on a missing ring entry at `last_settled_epoch` also moved inside the purchase branch — a top-up never needs those entries. Two tests cover it: `--amount 10` accepted for an active seat after the 43 -> 10 decrease, and the rejection message asserted not to contain the purchase wording.

**`shreds price` degrading.** `last_settled_epoch` is now `Option<u64>`; a missing or unparseable execution controller leaves Instant Price as `-` for every row, prints a one-line note saying why, and the rest of the listing stands.

**Changelog.** Entries added to both `crates/solana-cli` and `crates/solana-sdk` under `[Unreleased]`, and I removed the `skip-changelog` label that was masking the check. Worth noting the check was reporting green on the label, not on entries — so it wouldn't have caught this.

Re-verified live on DZ testnet (`-u devnet`): the epoch has since rolled past the reprice, and `lax-dz001` now reads `epoch_price` 10 / `instant_allocation_price` 10 — converged, as expected. The divergent 10 vs 43 reading from the original description is no longer reproducible on that device, so the incident case is covered by the unit test rather than a live reading.

crates/solana-cli/src/command/shreds/pay.rs
```diff
@@ -244,11 +315,18 @@ impl PayCommand {
         let (escrow_key, escrow_bump) =
             state::find_payment_escrow_address(&client_seat_key, &wallet_key);
         let (program_config_key, _) = state::find_program_config_address();
+        let (execution_controller_key, _) = state::find_execution_controller_address();
 
-        // Check which accounts already exist on-chain.
+        // Check which accounts already exist on-chain, and read the epoch the
```
nikw9944: Still says "on-chain" (also line 453); repo CLAUDE.md: "Always write 'onchain', never 'on-chain'." Carried over from round 1.

crates/solana-cli/src/command/shreds/pay.rs
```diff
@@ -102,6 +102,77 @@ fn epoch_warning_prompt(input: &EpochWarningInput) -> Option<String> {
     ))
 }
 
+/// Inputs for the `--amount` floor, separated from I/O for testability.
```
nikw9944: `SeatPriceInput` still has a `///` docstring; repo CLAUDE.md forbids `///` rustdoc on internal structs — switch to `//` (the `//` field comments below it already follow the rule). Carried over from round 1.

nikw9944 (dismissed):
No changes since the last re-check (head still bf43d77). All three substantive round-1 findings remain verified fixed: the settled-epoch floor and ring-presence requirements are scoped to the instant-allocation path with top-ups floored at the newest entry (tests pin both messages), `shreds price` degrades the Instant Price column to `-` instead of aborting when the execution controller is missing, and both crates have `[Unreleased]` changelog entries with CI fully green. Two Low convention nits remain unaddressed.

ben-dz:
Both nits fixed in 976ff33: `onchain` spelling in the three `pay.rs` comments (including the pre-existing line 445 you flagged), and `SeatPriceInput`'s `///` switched to `//` with the content kept. No behavior change.

martinsander00 (approved):

PR #406, opened 2026-07-31
shreds: add metro-dump to dump onchain metro price history

elitegreg:
## Problem

The `MetroHistory` account holds a 32-entry ring buffer of per-epoch USDC prices, but nothing in the CLI can read past the newest entry. The SDK exposes only:

- `parse_metro_history` — newest entry only (used by `shreds price`)
- `parse_metro_history_price_at_epoch` — one specific epoch (used by the `shreds pay` preflight)

Nothing reads the flags word either. So answering "what has this metro charged over the last few epochs" or "is this metro retransmit-only" means an explorer and a hex dump.

## Change

`parse_metro_history_account` in `crates/solana-sdk/src/shred_subscription/state.rs` verifies the discriminator, reads the flags and device count, and walks the whole ring newest-first. The walk is bounded by `total_count` rather than capacity for the same reason `find_ring_buffer_entry_offset` is: scanning all 32 slots would surface zero-initialized slots as real entries at epoch 0.

`shreds metro-dump` sits on top of it:

```
doublezero-solana shreds metro-dump [--metro <PUBKEY_OR_CODE>]
                                    [--epoch <N> | --num-epochs <N>]
                                    [--wide] [--json]
```

`--metro` accepts either the exchange public key or its code (`xlax`), resolved case-insensitively. `--epoch` and `--num-epochs` are mutually exclusive. Output is one row per (metro, epoch), sorted by metro code then epoch descending, as a markdown table or `--json`; public keys and ring internals sit behind `--wide`.

```
| Metro Code | Metro Name  | Epoch | Price (USDC) | Finalized | Retransmit Only | Devices |
|------------|-------------|-------|--------------|-----------|-----------------|---------|
| xlax       | Los Angeles | 1112  | 10           | yes       | yes             | 1       |
| xlax       | Los Angeles | 1111  | 10           | -         | yes             | 1       |
| xlax       | Los Angeles | 1108  | 43           | -         | yes             | 1       |
```

## Two behaviors worth a look in review

**`Finalized` is populated only on each metro's newest row.** `is_current_price_finalized` is account-level and describes the *current* price; repeating `yes` down 32 rows of history would imply state that is not recorded, so older rows show `-`.

**The exchange join is best effort.** Metro codes and names come from Serviceability on DZ Ledger. An unreachable ledger — or localnet, where `serviceability_program_id` has nothing to point at — degrades those two columns to `?` and prints a warning rather than failing, since the substance of the dump lives on Solana. Resolving `--metro` by *code* is the exception: that genuinely cannot be done without the listing, so it errors.

Also: `getProgramAccounts` returns the PDA but not its bump, so the bump comes from re-deriving `find_metro_history_address`. Comparing that derivation against the key the RPC returned is free, so the command skips (with a warning) any account whose `exchange_key` does not derive its own address.

## Testing

Nine new unit tests in `state.rs`, reusing the existing `metro_history_data` fixture (now also writing the discriminator, since the new parser checks it): ring walk order, wrapping past slot 0, the `total_count` bound, full 32-slot capacity, an uninitialized ring, header fields, both flag bits independently plus an unrelated bit, wrong discriminator, and short buffers.

Verified against the live testnet program on Solana devnet:

- All 8 metros' newest entries match `shreds price --all`'s `Base Price (USDC)` column exactly (xams 31, xfra 36, xlax 10, xlon 56, xnyc 87, xprg 46, xsin 30, xtyo 42) — the two read the same ring entry through different code paths.
- `--metro xlax` and `--metro 3SpDvEQ83VgeN67xygxSrgm2YyztNbR7NnHNH2qi7K4T` return byte-identical JSON (18 entries); `--metro XLAX` resolves the same.
- `--epoch 1108` returns 43 for xlax, matching that row in the full dump; a gap epoch returns no rows (`[]` under `--json`).
- A pubkey with no history PDA errors with `no metro history account at <PDA> for exchange <key>`.

`just fmt-check`, `cargo clippy --all-features --all-targets -- -Dclippy::all`, and the `doublezero-solana-sdk` + `doublezero-solana-cli` test suites (224 tests) all pass.


martinsander00:
Is this command intended for user use or for admin use? 

If it's the latter then we can use this cmd and modify it: https://github.com/malbeclabs/doublezero-shreds/blob/main/crates/onchain-admin/src/command/show/metro_history.rs

If it's the former then I think we should not call it metro-dump. Do you have thoughts on that? I will review the code more in depth tmrw but that's my first thought

PR #407, opened 2026-08-04
shreds pay: confirm the fastshreds.com transition before paying

kevin-dz:
## Summary of Changes

Adds the fastshreds.com transition notice to `shreds pay` with a y/N gate so users see the notification before paying. Implements malbeclabs/infra#2164 (cross-repo, so it will need closing by hand).

The notice block sits at `pay.rs:274-298`, entirely above `build_wallet` at line 301, so declining needs neither a loadable keypair nor a reachable cluster, and signs nothing. `command/mod.rs` is untouched — byte-identical to `main` — so the other callers of the shared prompt are unaffected by construction.

How the gate behaves:

1. **Both stdin and stdout must be a terminal.** The prompt is written to stdout, so with stdout redirected the question would land in the file while `read_line` blocked on a terminal showing nothing. Anything short of a full terminal falls through to print-only, which is the behavior already settled for the non-interactive case. Dry runs also print without prompting, matching the epoch-remaining prompt, since a simulation signs and sends nothing. `--accept-deprecation-notice` skips the prompt for interactive batch and multi-seat workflows; the notice still prints.
2. **Declining bails, so it exits non-zero**, consistent with declining the epoch-remaining prompt in the same command, and so `shreds pay ... && notify` does not fire on a decline.

The stronger reason for routing non-terminal stdin to print-only is that keypair-from-stdin (`solana-client-tools/src/keypair/loader.rs:49-52`) requires a non-terminal stdin, so prompting there would consume the keypair bytes.

CHANGELOG.md updated. No metrics exposed.

### Copy

Settled with the issue author, so nothing here is open.

The `apt` upgrade instruction is dropped. The release ships deb, rpm and tar.gz (`release/.goreleaser.doublezero-solana-cli.yaml`) and `scripts/install-doublezero-solana.sh` auto-detects all three, so naming one package manager would have been wrong for rpm and tarball users. The abort message points at fastshreds.com instead, which is where an operator who declines is usually headed.

The prompt keeps the shared helper's `Proceed? [y/N]` — a bare Enter counts as no — rather than the issue's `Would you like to continue? y/n`, for consistency with the epoch-remaining prompt in the same command. The notice's closing sentence was reworded to "This command will fund your seat now" so the appended question attaches to the payment rather than to closing the account.

### Out of scope, confirmed in review

`shreds withdraw` carries no notice — the assumption is that it gets its own message once `shreds pay` is disabled. And raising the onchain `min_compatible_cli_version` in `malbeclabs/doublezero-shreds`, which is the only lever that reaches operators who never upgrade, should follow this shipping rather than precede it.

## Testing Verification

**Mainnet-beta, end to end.** Declined at the prompt, dry-ran, then bought and withdrew a real seat in the same sitting with a throwaway keypair. Confirms what unit tests cannot: `y` carries through to a landed transaction, and the notice does not alter the payment path. Note this predates the dry-run change above, so the dry run in that session did prompt.

**Terminal matrix**, with no keypair present and the RPC on a dead port. Verified across stdin/stdout as terminal, file and pipe, plus `--dry-run` and `--accept-deprecation-notice`: the notice prints in all eight combinations, and the prompt fires only for a full terminal with no skip flag. Invalid input re-prompts; `n` gives `Error: Aborted. Manage your account at https://www.fastshreds.com.` and exit 1; `y` proceeds into `build_wallet`.

**Checks.** 451 workspace tests pass; `cargo fmt --check` clean workspace-wide. Local clippy cannot go green because `main` itself fails `-Dclippy::all` under rustc 1.95 (2 lints in `contributor-rewards`, 2 in `solana-cli`, none in files this diff touches, all confirmed present on unmodified `main`). The repo pins 1.92.0, so CI is the authority; `pay.rs` draws zero lints.


juan-malbeclabs (commented):
This is the first code review we're doing in this page, so a note on scope before the findings: I read the diff against `main`, checked the two existing callers of the prompt helper, and traced the ordering claims in the PR body. The placement claim holds — the notice block sits at `pay.rs:294-314`, entirely above `build_wallet` at line 317, and the `command/mod.rs` split preserves existing callers' behavior exactly (`Some('y') => true`, `Some('n') | None => false`, invalid input re-prompts). `convert_2z.rs:63`, `validator_deposit.rs:204` and the epoch-remaining prompt at `pay.rs:482` are unaffected. Nothing is signed or sent on the decline path, no new network calls, no new dependencies.

Going forward I'd like reviews here to focus on correctness first, then consistency within the command being touched, then coverage. That's the order below.

## Correctness

**1. `> log.txt` now hangs every interactive run** (`pay.rs:297-299`). The gate checks only `stdin().is_terminal()`, but the prompt is written to `out`. With stdout redirected and stdin still a tty — `doublezero-solana shreds pay ... > pay.log`, or piping through `tee`/`less` — the prompt lands in the file and the process blocks on `read_line` against a blank terminal. This hazard already existed for the epoch prompt, but that one only fires in the last 10% of an epoch; this one fires on *every* invocation, which makes it a materially different exposure. Suggest requiring both ends to be a terminal:

```rust
let should_prompt = !self.accept_deprecation_notice
    && std::io::stdin().is_terminal()
    && std::io::stdout().is_terminal();
```

Redirected stdout then falls into the print-only path, which is the behavior already decided to be correct for the non-interactive case.

**2. `--dry-run` prompts, but the epoch warning deliberately doesn't.** `EpochWarningInput.dry_run` (`pay.rs:474`) exists precisely so a simulation isn't gated behind a y/N. A dry run signs and sends nothing, so gating it is a prompt with nothing behind it — and it's inconsistent inside the same command. `self.write_opts.dry_run` is readable before `build_wallet`, so it's one more term: `&& !self.write_opts.dry_run`. I'd print-only on dry-run.

**3. Exit code on decline (open question 1) — recommend bailing rather than exiting 0.** Two reasons beyond the cosmetic `Error:` prefix:

- Within a single command, declining the epoch prompt exits non-zero while declining this one exits 0. An operator can't reason about that.
- `shreds pay ... && notify "seat purchased"` fires on a decline. Exit 0 asserts the command did what it was asked, and no payment was made. That's a worse signal than an `Error:` prefix on a line the operator caused themselves.

If we take this, the whole `command/mod.rs` split reverts and the diff shrinks to `pay.rs` alone: `try_prompt_proceed_confirmation(out, DEPRECATION_NOTICE, "Aborted. Upgrade with ...")`. Keeping exit 0 is defensible, but then the epoch decline should arguably match it.

**4. Worth calling out as correct-by-design:** because keypair-from-stdin (`solana-client-tools/src/keypair/loader.rs:49-52`) requires a non-tty stdin, and non-tty routes to print-only, the prompt never consumes the keypair bytes. That's a real hazard the non-terminal decision avoids, and it's a stronger argument for that decision than the cron/ansible one.

## Style and consistency

- **`NoticeDelivery` + `notice_delivery` is more machinery than the predicate deserves.** Twelve lines of enum and function plus a four-case test to express `accepted || !is_terminal`. Our single-caller-helper rule does exempt units under test, but the test is tautological with the implementation — it restates the boolean table rather than pinning behavior. I'd collapse to a local `let should_prompt = ...;` at the call site (see item 1) and drop the enum along with `test_notice_delivery_only_prompts_an_unaccepted_terminal`. That also turns items 1 and 2 into a one-line change instead of widening a function signature to four parameters.
- `test_notice_text_has_no_trailing_period` is worth keeping — it guards a real coupling, since `try_prompt_proceed` appends `". Proceed? [y/N]"`. It survives the collapse above.
- `⚠️  {DEPRECATION_NOTICE}.` at `pay.rs:313` duplicates the helper's own prefix-and-period formatting. Not worth extracting, but the two can drift silently if the helper's format changes.
- **CHANGELOG blank line** at `CHANGELOG.md:12`: it separates the two new entries from the rest of the list, which makes the whole list loose in rendered markdown — every item picks up paragraph spacing, not just the new ones. Drop it. The `(malbeclabs/infra#2164)` cross-repo reference is consistent with the existing precedent on line 18, so no change needed there.
- The string-continuation constant is correct: every line breaks after a space, `\<newline>` strips the following indentation, and the joined text has no stray whitespace and no `\n`.

## Test coverage

Only the pure predicate and the constant are covered. The two branches inside `execute` — the abort message text and the print-only `writeln!` — are verified by hand only, so a typo in either user-visible string ships silently. Not blocking, since they're unreachable in a unit test without splitting `execute`, but the manual matrix in the PR description isn't regression protection. I'd leave it as-is rather than add a weak assertion on the constant's contents.

## Smaller notes

- `Aborted. Upgrade with \`sudo apt update && sudo apt install doublezero-solana\`` is a slight non sequitur for an operator who declined *because* they intend to use the web front end. Something closer to `Aborted. Manage your account at www.fastshreds.com.` reads truer to why someone answers `n`. Related: the apt-only instruction won't help anyone who installed by another route.
- Invalid input re-prompts by reprinting the entire ~450-character notice. Pre-existing in the helper, but newly noticeable at this message length.
- Open questions 3 (`shreds withdraw` carries no notice) and 4 (`min_compatible_cli_version` as the lever for operators who never upgrade) both read correct, and are properly out of scope here.
- Open question 5: I'd leave `--accept-deprecation-notice` out of the notice text. Automation never sees the prompt, and advertising the skip flag to interactive users undercuts the point of the transition.

## Verdict

Sound and well-scoped, and the payment path is genuinely untouched. I'd land it after fixing the redirected-stdout hang (item 1) — that one bites real operators on every run. The dry-run gate (2), exit-code consistency (3), and the `notice_delivery` collapse are best taken together, since they touch the same six lines.

crates/solana-cli/src/command/shreds/pay.rs
```diff
@@ -254,6 +291,28 @@ impl PayCommand {
         ctx: &CliContext,
         out: &mut impl Write,
     ) -> Result<()> {
+        // Before the wallet is built and before any RPC: an operator who
+        // declines should not need a loadable keypair or a reachable cluster,
+        // and nothing should be signed or sent on their behalf.
+        match notice_delivery(
+            self.accept_deprecation_notice,
+            std::io::stdin().is_terminal(),
```
juan-malbeclabs: **Needs a change: this hangs any run with redirected stdout.**

The gate checks only `stdin().is_terminal()`, but the prompt is written to `out`. With stdout redirected and stdin still a tty — `doublezero-solana shreds pay ... > pay.log`, or piping through `tee`/`less` — the prompt lands in the file and the process blocks on `read_line` against a blank terminal. The hazard already existed for the epoch prompt, but that one only fires in the last 10% of an epoch; this fires on every invocation, so the exposure is materially different.

Same place, second term: `--dry-run` currently prompts, while the epoch warning deliberately skips on dry-run (`EpochWarningInput.dry_run`, line 474). A dry run signs and sends nothing, so the prompt gates nothing — and it's inconsistent inside the same command. `self.write_opts.dry_run` is readable here, before `build_wallet`.

Both fold into one predicate:

```rust
let should_prompt = !self.accept_deprecation_notice
    && !self.write_opts.dry_run
    && std::io::stdin().is_terminal()
    && std::io::stdout().is_terminal();
```

Redirected stdout and dry-run then fall into the print-only path, which is already the behavior we decided is right for the non-interactive case.

crates/solana-cli/src/command/shreds/pay.rs
```diff
@@ -65,6 +78,26 @@ fn is_seat_already_active(seat_data: Option<&[u8]>) -> bool {
...
+}
+
+/// Decide how to deliver the deprecation notice. Separated from I/O for testability.
+fn notice_delivery(accepted: bool, stdin_is_terminal: bool) -> NoticeDelivery {
+    if accepted || !stdin_is_terminal {
+        NoticeDelivery::PrintOnly
+    } else {
+        NoticeDelivery::Prompt
+    }
+}
```
juan-malbeclabs: **Suggest collapsing this to a local bool at the call site.**

Twelve lines of enum plus function to express `accepted || !is_terminal`. Our single-caller-helper rule does exempt a helper that is itself the unit under test, but the test here restates the boolean table rather than pinning behavior, so the exemption isn't really earned.

Dropping the enum and inlining `let should_prompt = ...;` also turns the two fixes in my other comment into a one-line change, instead of widening this signature to four parameters and threading `dry_run` and a second `is_terminal` through it.

crates/solana-cli/src/command/shreds/pay.rs
```diff
@@ -254,6 +291,28 @@ impl PayCommand {
...
+        // and nothing should be signed or sent on their behalf.
+        match notice_delivery(
+            self.accept_deprecation_notice,
+            std::io::stdin().is_terminal(),
+        ) {
+            NoticeDelivery::Prompt => {
+                // A decline is a normal outcome, not a failure, so it prints
+                // plainly and exits 0 instead of returning an `Err` that `main`
+                // would render as "Error: Aborted...".
+                if !crate::command::try_prompt_proceed(out, DEPRECATION_NOTICE)? {
```
juan-malbeclabs: **Open question 1 — I'd bail here rather than print and return `Ok(())`.** Two reasons beyond the cosmetic `Error:` prefix:

- Within this one command, declining the epoch prompt at line 482 exits non-zero while declining here exits 0. An operator can't reason about that.
- `shreds pay ... && notify "seat purchased"` fires on a decline. Exit 0 asserts the command did what it was asked, and no payment was made — a worse signal than an `Error:` prefix on a line the operator caused themselves.

If we take this, the whole `command/mod.rs` split reverts and this becomes `try_prompt_proceed_confirmation(out, DEPRECATION_NOTICE, "Aborted. ...")?`. Keeping exit 0 is defensible, but then the epoch decline should arguably match it — what I'd like to avoid is the two prompts disagreeing.

crates/solana-cli/src/command/shreds/pay.rs
```diff
@@ -254,6 +291,28 @@ impl PayCommand {
...
+                    writeln!(
+                        out,
+                        "Aborted. Upgrade with `sudo apt update && sudo apt install doublezero-solana`"
```
juan-malbeclabs: Wording: `Upgrade with sudo apt ...` is a slight non sequitur for the operator who declined *because* they intend to use the web front end — the upgrade instruction is what they just said no to. Something closer to `Aborted. Manage your account at www.fastshreds.com.` reads truer to why someone answers `n`. Related: the apt-only instruction doesn't help anyone who installed by another route.

crates/solana-cli/src/command/shreds/pay.rs
```diff
@@ -254,6 +291,28 @@ impl PayCommand {
...
+                    )?;
+                    return Ok(());
+                }
+            }
+            NoticeDelivery::PrintOnly => writeln!(out, "⚠️  {DEPRECATION_NOTICE}.")?,
```
juan-malbeclabs: Nit: this duplicates the `⚠️  {}. ` prefix-and-period formatting that `try_prompt_proceed` applies to the same constant. Not worth extracting for two call sites, but the two will drift silently if the helper's format ever changes — and `test_notice_text_has_no_trailing_period` only guards the helper's half of that coupling.

crates/solana-cli/src/command/shreds/pay.rs
```diff
@@ -660,6 +719,26 @@ mod tests {
...
+
+    #[test]
+    fn test_notice_delivery_only_prompts_an_unaccepted_terminal() {
+        assert_eq!(notice_delivery(false, true), NoticeDelivery::Prompt);
+        // Cron, ansible and piped stdin would otherwise read EOF as "no" and
+        // abort a payment the operator never got to see.
+        assert_eq!(notice_delivery(false, false), NoticeDelivery::PrintOnly);
+        assert_eq!(notice_delivery(true, true), NoticeDelivery::PrintOnly);
+        assert_eq!(notice_delivery(true, false), NoticeDelivery::PrintOnly);
+    }
```
juan-malbeclabs: This test is tautological with the implementation — it restates `accepted || !stdin_is_terminal` as a four-row table, so it can only fail if someone inverts the expression, and it would need editing (not just re-running) for either fix in my other comments. I'd drop it along with the enum.

`test_notice_text_has_no_trailing_period` below is worth keeping: it guards a real coupling, since `try_prompt_proceed` appends `". Proceed? [y/N]"`. It survives the collapse.

crates/solana-cli/CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- `command`: split the shared interactive prompt into `try_prompt_proceed`, which returns whether the operator chose to proceed, and `try_prompt_proceed_confirmation`, which is now a thin wrapper that bails with the caller's abort message. Existing callers keep their behavior exactly; the new form is for callers that want to report a decline themselves without the `Error:` prefix (malbeclabs/infra#2164)
+
```
juan-malbeclabs: Drop this blank line. It splits the `[Unreleased]` list into two markdown lists, which makes the whole thing render loose — every item on the page picks up paragraph spacing, not just the two new ones.

The `(malbeclabs/infra#2164)` cross-repo reference is fine, matching the existing precedent on line 18.

crates/solana-cli/src/command/mod.rs
```diff
@@ -132,6 +132,18 @@ pub(crate) fn try_prompt_proceed_confirmation(
...
+    } else {
+        anyhow::bail!("{abort_message}")
+    }
+}
+
+// Returns whether the operator chose to proceed. Callers that treat a decline
+// as an error should use `try_prompt_proceed_confirmation` instead; this form is
+// for callers that want to report the decline themselves without the `Error:`
+// prefix `main` puts on a returned `Err`.
+pub(crate) fn try_prompt_proceed(out: &mut impl Write, prompt_message: &str) -> Result<bool> {
```
juan-malbeclabs: No objection to the split as written — existing callers keep their behavior exactly (`Some('y') => true`, `Some('n') | None => false`, invalid input re-prompts), and `convert_2z.rs:63` / `validator_deposit.rs:204` are unaffected.

Flagging only that this whole hunk exists to serve the exit-0 decision in `pay.rs:305`. If we go with bailing there, the split has no caller and this file drops out of the diff entirely.

juan-malbeclabs (commented):
Inline notes for the specific spots that need a change, following up on the review above. Ordered by impact: the stdout-tty gate is the one I'd consider blocking; the rest are consistency and cleanup.

kevin-dz:
All four correctness items taken, in 1b42b66.

**1. Redirected stdout.** Fixed as suggested — `should_prompt` requires both ends to be a terminal. Verified across five stdin/stdout combinations: `terminal/terminal` prompts, `terminal/file`, `terminal/pipe`, `file/terminal` and `file/file` all print and proceed with no hang.

**2. Dry-run.** Fixed, `&& !self.write_opts.dry_run`. One consequence worth naming: `--dry-run` no longer exercises the prompt at all, so the mainnet dry run in the testing section predates this and did prompt.

**3. Exit code.** Taken. `command/mod.rs` is reverted to byte-identical with `main`, so the split is gone and this is confined to `pay.rs`. The `&& notify` argument is what convinced me — that's a concrete failure my "nothing failed" reasoning didn't account for.

**4. Keypair-from-stdin.** Good catch, and you're right that it's the stronger argument. It's now the primary rationale in the code comment and the CHANGELOG, replacing the cron/ansible one.

**Collapse.** Done — `NoticeDelivery` and `notice_delivery` deleted (−20 lines) along with the tautological test (−10). Your item 1 patch was already the collapse, so items 1, 2 and this landed as one edit to one block. Kept `test_notice_text_has_no_trailing_period`.

**CHANGELOG.** Blank line dropped, list is tight again.

crates/solana-cli/src/command/shreds/pay.rs
kevin-dz: Both terms taken exactly as written — the predicate is identical to your snippet. Verified across terminal/file/pipe on both ends plus dry-run: the notice always prints, the prompt fires only on a full terminal.

crates/solana-cli/src/command/shreds/pay.rs
kevin-dz: Collapsed. Enum and function are gone, with no references left.

crates/solana-cli/src/command/shreds/pay.rs
kevin-dz: Bailing now, so both declines in this file exit non-zero with the same 'Aborted. <next step>.' shape.

crates/solana-cli/src/command/shreds/pay.rs
kevin-dz: Reworded to 'Aborted. Manage your account at https://www.fastshreds.com.' Your non-apt point applied to the notice too.

crates/solana-cli/src/command/shreds/pay.rs
kevin-dz: Leaving as-is per your call. Noted that the kept test only covers the helper's half of the coupling.

crates/solana-cli/src/command/shreds/pay.rs
kevin-dz: Dropped along with the enum. Kept test_notice_text_has_no_trailing_period.

crates/solana-cli/CHANGELOG.md
kevin-dz: Dropped, the list renders tight again.

crates/solana-cli/src/command/mod.rs
kevin-dz: Went with bailing, so this hunk is gone — mod.rs is byte-identical to main and out of the diff entirely.

juan-malbeclabs (dismissed):
All eight inline comments and every item in the review body are addressed. Verified against the code rather than the replies.

**Correctness.** The predicate at `pay.rs:283-286` is identical to the suggested one, both ends of the terminal plus `!dry_run`. The dry-run term is the same value the epoch prompt uses — `wallet.dry_run` comes from `opts.dry_run` in `mod.rs:68` — so the two prompts in this command cannot disagree on what a simulation is. Declining now routes through `try_prompt_proceed_confirmation`, which bails at `mod.rs:151`, so it exits non-zero and `&& notify` no longer fires. `command/mod.rs` is byte-identical to `main`: the split reverted entirely and the file is out of the diff. Keypair-from-stdin is the primary rationale in both the code comment and the CHANGELOG.

**Style.** `NoticeDelivery` and `notice_delivery` are gone with no references left anywhere in `crates/`, collapsed to a local `should_prompt` at the call site. The tautological test is dropped and `test_notice_text_has_no_trailing_period` survives. The CHANGELOG list has no internal blank line. The abort message points at fastshreds.com, and the non-apt point was applied to the notice text too.

**Scope.** Two files, +53/-1. The whole notice block is `pay.rs:273-296`, above `build_wallet` at 299, so the payment path is untouched. `cargo fmt --check` clean, crate tests pass, all five CI checks green.

Two things to carry forward, neither blocking and both already named in the PR: `--dry-run` no longer exercises the prompt, so the mainnet dry run in the testing section predates that change and did prompt; and invalid input still reprints the full notice, which is the helper's loop and a fix for the helper if it ever becomes annoying.

Good to land.

crates/solana-cli/src/command/shreds/pay.rs
```diff
@@ -660,6 +701,16 @@ mod tests {
...
+    // --- deprecation notice tests ---
+
+    #[test]
+    fn test_notice_text_has_no_trailing_period() {
+        // `try_prompt_proceed_confirmation` appends ". Proceed? [y/N]", and
+        // formats the message as a single `writeln!`.
+        assert!(!DEPRECATION_NOTICE.ends_with('.'));
+        assert!(!DEPRECATION_NOTICE.contains('\n'));
+    }
+
```
martinsander00, line 710: let's add 2 tests: one that asserts that when the user doesn't accept the flag we print this thing: `writeln!(out, "⚠️  {DEPRECATION_NOTICE}.")?;` + the `DEPRECATION_NOTICE`, and one that asserts that when the user does accept the flag that they can go through the flow normally.

kevin-dz: Both added

crates/solana-cli/src/command/shreds/pay.rs
```diff
@@ -660,6 +700,84 @@ mod tests {
...
+        let mut out = Vec::new();
+
+        let result = pay_command(&[]).execute(None, &ctx, &mut out).await;
+
+        assert_eq!(
+            String::from_utf8(out).expect("output is utf8"),
+            format!("⚠️  {DEPRECATION_NOTICE}.\n")
+        );
+        let error = result.unwrap_err().to_string();
+        assert!(
```
martinsander00: can you make this assert! -> assert_eq! so we pin the exact line + log instead of just checking contains?

crates/solana-cli/src/command/shreds/pay.rs
```diff
@@ -660,6 +700,84 @@ mod tests {
...
+            .execute(None, &ctx, &mut out)
+            .await;
+
+        assert_eq!(
+            String::from_utf8(out).expect("output is utf8"),
+            format!("⚠️  {DEPRECATION_NOTICE}.\n")
+        );
+
+        let error = result.unwrap_err().to_string();
+        assert!(
```
martinsander00: same here

crates/solana-cli/src/command/shreds/pay.rs
kevin-dz: Done for both tests

martinsander00 (approved):

PR #409, opened 2026-08-14
solana-client-tools: add a payload budget and a checked encoder for Squads vaults

### LINKED ISSUE #4184: solana-client-tools: add Squads vault transaction bundling to the squads module
Target repo: [doublezerofoundation/doublezero-offchain](https://github.com/doublezerofoundation/doublezero-offchain)

## Summary

The `squads` module encodes one payload through `encode_vault_transaction` and prints it. Nothing accounts for what Squads wraps around that payload, and nothing refuses a payload that will not fit, so every consumer has to redo the accounting. `claim` and `set-proportion` are queued behind this.

- A payload's budget is `MAX_TRANSACTION_SIZE - 384 - instruction_count`. The transaction that has to carry a payload is the Squads create transaction, not the payload itself, and a payload that overruns it fails after a human has already imported it and collected approvals.
- The 384 is derived, and rounded up from 381:
  - 286 for the `vault_transaction_create` envelope: one signature, the header, five accounts (multisig, transaction account, creator, rent payer, system program) plus the Squads program id, the blockhash, and an instruction whose data carries the 8-byte discriminator, `vault_index`, `ephemeral_signers`, the 4-byte length prefix, and `memo: None`.
  - 52 for the compute budget pair the app bundles: the compute budget program key, a 5-byte limit payload, and a 9-byte price payload.
  - 57 for a bundled `proposal_create` (the proposal account and its instruction) and 15 for a bundled `proposal_approve`, which is how one transaction takes a payload all the way to the approvers.
  - 2 in case the app compiles that transaction as v0 rather than legacy.
  - Minus 31, which is what Squads' `TransactionMessage` encoding saves against a legacy message: the 32 bytes of blockhash it does not carry, less the 1 byte of `address_table_lookups` length it always writes.
  - Plus 1 per payload instruction, because `data: SmallVec<u16, u8>` spends 2 bytes on a length a legacy message writes in 1. That stops holding once an instruction's data reaches 128 bytes, where the legacy prefix also grows to 2, so the term is conservative rather than exact.
- `vault_transaction_execute` costs `277 + 33n` for `n` payload account keys, because by then the payload's instruction data lives in the transaction account and does not travel. It is looser than create at every size, so create binds. Check both anyway, so nothing can certify a payload that imports and then cannot execute.
- A budget accessor, for a caller that grows a single instruction until it stops fitting. `claim` appends `ClaimHoldingId`s.
- A checked encoder, which refuses an oversized payload instead of printing one. `set-proportion` emits a fixed single instruction.
- Tests need no cluster.
  - Fixed pubkeys and hand-computed lengths, including the per-instruction term at an instruction whose data crosses 128 bytes
  - A payload one byte inside the budget and one byte over it
  - The budget the accessor reports against the length the checked encoder measures, so the two cannot drift apart

karl-dz:
Closes https://github.com/malbeclabs/doublezero/issues/4184

## Summary
- Add `vault_transaction_payload_budget`, reporting the bytes a payload of a given instruction count may serialize to, for a caller that grows a single instruction until it stops fitting
- Add `try_encode_vault_transaction`, which measures the payload and refuses one that cannot work. `print_vault_transaction` becomes `try_print_vault_transaction` and routes through it
- Derive the reserve as const arithmetic rather than in a comment, one term per line, from the `vault_transaction_create` envelope plus the compute budget pair, `proposal_create`, and `proposal_approve` the Squads app bundles into that same transaction. Compile-time assertions pin each block's subtotal and the 384 total, so a changed term fails the build instead of leaving a stale literal. A comment records the three app behaviors the derivation assumes away and what each costs
- Refuse a payload naming a signer other than the vault, and one carrying more than 48 instructions. Both import and collect approvals before failing at execute, unlike an oversized payload, which fails at import
- Check the `vault_transaction_execute` transaction as well, at 329 bytes plus 33 per payload account key. The issue specifies 277, which omits the compute budget pair the app bundles there too
- Record the conventions this change was written under in `CLAUDE.md`, covering numeric literals, prose in comments and documentation, the `try_` prefix, and where workspace dependency features belong

## Testing
- Hand-computed payload lengths at 3, 127 and 128 instruction-data bytes, spanning the legacy length prefix boundary
- A payload landing exactly on the budget and one byte over it, at one instruction and at two
- The wrapper Squads builds around a budget-sized payload, assembled and measured at 1,229 bytes of the 1,232 available
- The widest payload the budget accepts, at 24 account keys and 1,121 bytes to execute
- Refusals for an empty payload, a foreign signer, and 49 instructions
- A downstream caller's vault print path against devnet, run before and after the change, byte-identical

### DIFF
--- a/crates/solana-client-tools/src/squads.rs
+++ b/crates/solana-client-tools/src/squads.rs
@@ -5,13 +5,117 @@ use solana_loader_v3_interface::state::UpgradeableLoaderState;
 use solana_message::Message;
 use solana_sdk::{instruction::Instruction, pubkey, pubkey::Pubkey};
 
-use crate::rpc::{NetworkEnvironment, SolanaConnection};
+use crate::{
+    rpc::{NetworkEnvironment, SolanaConnection},
+    transaction::MAX_TRANSACTION_SIZE,
+};
 
 // Squads Protocol v4 multisig program.
 pub const SQUADS_V4_PROGRAM_ID: Pubkey = pubkey!("SQDS4ep65T869zMMBKyuUq6aD6EgTu8psMjkvj52pCf");
 
 const VAULT_SEED_PREFIX: &[u8] = b"multisig";
 const VAULT_SEED: &[u8] = b"vault";
+
+// A payload travels as the `transaction_message` of `vault_transaction_create`, and the
+// app bundles the compute budget pair, a `proposal_create`, and a `proposal_approve` into
+// that same transaction, which is how one transaction takes a payload all the way to the
+// approvers. That transaction is what has to fit MAX_TRANSACTION_SIZE, so what it spends
+// is what a payload does not get to spend. Each block below adds up to the byte count its
+// assertion names, and the total is the reserve.
+
+// vault_transaction_create, as a legacy transaction.
+const VAULT_TRANSACTION_CREATE_BYTES: usize = 1 // signature count (shortvec)
+    + 64 // creator signature
+    + 3 // message header
+    + 1 // account key count (shortvec)
+    + 32 // multisig key
+    + 32 // transaction account key
+    + 32 // member key, both creator and rent payer
+    + 32 // system program key
+    + 32 // Squads program key
+    + 32 // recent blockhash
+    + 1 // instruction count (shortvec)
+    + 1 // program id index
+    + 1 // account index count (shortvec)
+    + 5 // account indexes
+    // Instruction data length (shortvec), 2 bytes once a payload pushes the data past 127.
+    + 2
+    + 8 // instruction discriminator
+    + 1 // vault_index
+    + 1 // ephemeral_signers
+    + 4 // transaction_message length (borsh Vec<u8> prefix)
+    + 1; // memo: None
+const _: () = assert!(VAULT_TRANSACTION_CREATE_BYTES == 286);
+
+// The compute budget pair the app bundles.
+const COMPUTE_BUDGET_PAIR_BYTES: usize = 32 // compute budget program key
+    + 1 // program id index (limit)
+    + 1 // account index count (shortvec)
+    + 1 // instruction data length (shortvec)
+    + 5 // SetComputeUnitLimit payload, a discriminator and a u32
+    + 1 // program id index (price)
+    + 1 // account index count (shortvec)
+    + 1 // instruction data length (shortvec)
+    + 9; // SetComputeUnitPrice payload, a discriminator and a u64
+const _: () = assert!(COMPUTE_BUDGET_PAIR_BYTES == 52);
+
+// The proposal_create the app bundles, whose other four accounts are already keys of the
+// create instruction.
+const PROPOSAL_CREATE_BYTES: usize = 32 // proposal account key
+    + 1 // program id index
+    + 1 // account index count (shortvec)
+    + 5 // account indexes
+    + 1 // instruction data length (shortvec)
+    + 8 // instruction discriminator
+    + 8 // transaction_index, a u64
+    + 1; // draft
+const _: () = assert!(PROPOSAL_CREATE_BYTES == 57);
+
+// The proposal_approve the app bundles, all of whose accounts are already keys.
+const PROPOSAL_APPROVE_BYTES: usize = 1 // program id index
+    + 1 // account index count (shortvec)
+    + 3 // account indexes
+    + 1 // instruction data length (shortvec)
+    + 8 // instruction discriminator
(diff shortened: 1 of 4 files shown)


crates/solana-client-tools/src/squads.rs
```diff
@@ -5,14 +5,118 @@ use solana_loader_v3_interface::state::UpgradeableLoaderState;
...
+// What Squads spends around a payload, against the length that payload measures as a
+// legacy message.
+const VAULT_TRANSACTION_RESERVED_BYTES: usize = VAULT_TRANSACTION_CREATE_BYTES
+    + COMPUTE_BUDGET_PAIR_BYTES
+    + PROPOSAL_CREATE_BYTES
+    + PROPOSAL_APPROVE_BYTES
+    + 2 // in case the app compiles that transaction as v0 rather than legacy
+    - 32 // recent blockhash a Squads TransactionMessage does not carry
+    + 1 // address_table_lookups length a Squads TransactionMessage always writes
+    + 3; // rounding up the 381 of everything above
```
bgm-malbeclabs, line 92: Nit, not a blocker. The comment right below spells out that a memo spends from the payload's own headroom, so this is a known tradeoff and not a miss.

The part worth naming is what the caller sees. `try_encode_vault_transaction` accepts a payload at exactly the budget, then the operator types any memo at all in the import dialog and the create transaction goes over 1,232 bytes. Borsh charges 4 bytes for `Some(String)` over `None` before a single character of text, so these 3 bytes of rounding cover a zero-length memo minus one byte. The failure then arrives from the Squads app as a size error that names nothing the caller can act on.

Either reserve for a short memo, or say on `vault_transaction_payload_budget` that the budget assumes no memo. A caller sizing to the last byte should know what it is spending.

crates/solana-client-tools/src/squads.rs
```diff
@@ -5,14 +5,118 @@ use solana_loader_v3_interface::state::UpgradeableLoaderState;
...
+
+// Payload instructions run as separate invocations from vault_transaction_execute,
+// against a runtime instruction trace that holds 64 entries for the whole transaction.
+// Execute spends three of them on the compute budget pair and on itself, leaving around
+// 61, and fewer for every payload instruction that invokes further.
+//
+// 48 is an arbitrarily conservative lower bound, not that ceiling. Nothing here measures
+// how deep a payload's own invocations go, so the limit leaves room for them rather than
+// pricing them, and a real payload needing more is the reason to raise it.
+const MAX_PAYLOAD_INSTRUCTIONS: usize = 48;
```
bgm-malbeclabs: Nit, not a blocker. `vault_transaction_payload_budget` is public so a caller can size its work up front, but this cap is private. A caller batching small instructions can stay inside the byte budget, append a 49th instruction, and find the limit only in the error from `try_encode_vault_transaction`.

Making this `pub` gives the count side the same thing the byte side already offers.

bgm-malbeclabs (commented):
Reviewed the derivations by hand and ran the crate tests (17 squads tests pass) and `cargo clippy --all-targets` (clean).

What checks out: all five `const _: () = assert!(...)` blocks add up, and the Squads v4 account lists they encode are right (`VaultTransactionCreate` 5 accounts, `ProposalCreate` 5, `ProposalVote` 3, `VaultTransactionExecute` 4). The one byte per instruction charge is a correct upper bound, since Squads writes instruction data length as `SmallVec<u16, u8>` against legacy shortvec, and every other length prefix over-charges rather than under-charges. The claim that create binds before execute holds: create allows 24 zero-data account keys against execute's 27, and execute does not grow with instruction count, so the execute check really is unreachable. No callers are left on the removed `print_vault_transaction` and `encode_vault_transaction`.

No correctness bug found. Two nits below, neither a blocker.

crates/solana-client-tools/src/squads.rs
karl-dz: Documented rather than reserved, since the field is unbounded and a wider reserve moves the failure without removing it. How does this look? https://github.com/doublezerofoundation/doublezero-offchain/pull/409/commits/ed96c4d2f0bbc012b9aac1b2fe083bca0083ad50

crates/solana-client-tools/src/squads.rs
karl-dz: Agree. How does this look? https://github.com/doublezerofoundation/doublezero-offchain/pull/409/commits/646027dd177f8715dfecbd758140b372644c636f

bgm-malbeclabs (approved):

PR #410, opened 2026-08-14
solana-sdk: mirror ValidatorClientRewards as a zero-copy struct

### LINKED ISSUE #4183: solana-sdk: mirror ValidatorClientRewards as a zero-copy struct
Target repo: [doublezerofoundation/doublezero-offchain](https://github.com/doublezerofoundation/doublezero-offchain)

## Summary

`ValidatorClientRewards` is read through hand-written byte offsets. Its direct sibling `ValidatorPublisherRewards` is a mirrored `Pod` struct.

- Mirror the program layout, taking the derive list from `ValidatorPublisherRewards`. That is 176 bytes, so the account data is 184.

  ```rust
  pub struct ValidatorClientRewards {
      pub client_id: u16,
      pub bump_seed: u8,
      _padding_0: [u8; 5],
      pub manager_key: Pubkey,
      pub short_description_bytes: [u8; 64],
      pub claim_holding_count: u32,
      _padding_1: [u8; 4],
      _gap: StorageGap<2>,
  }
  ```

- Add the compile-time size assertion the other mirrors carry, so a layout change onchain fails the build rather than silently shifting a field.
- Carry a `checked_short_description` accessor matching the program, in place of the `Option<String>` the info struct returns.
- Delete the five `VCR_*_OFFSET` constants, `VCR_SHORT_DESCRIPTION_LEN`, `VCR_ACCOUNT_DATA_LEN`, the standalone discriminator constant, `parse_validator_client_rewards`, and `ValidatorClientRewardsInfo`. The discriminator constant's only consumers are that parser and `solana-fork`, and the mirror's `PrecomputedDiscriminator` implementation replaces it, so leaving it in place keeps a second copy of the discriminator string.
- Construction sites build with `Zeroable::zeroed()` and field assignment, since the padding and gap fields are private.
- `init-holding` reads no field, only asserting that the account is a `ValidatorClientRewards`, so it wants a discriminator check rather than a decode. It and `show` also hold the account data rather than the `Account`, so they decode through `checked_from_bytes_with_discriminator` rather than `ZeroCopyAccountOwnedData::from_account`.
- `solana-fork` builds a synthetic account by copying bytes to those offsets, and sizes its rent exemption from `VCR_ACCOUNT_DATA_LEN`. The rent call takes the built buffer's length instead. Delete its parse-back round trip and the three `ensure!` checks over it, which are tautological against a struct the same function just built.
- The `solana-sdk` changelog's unreleased section carries a bullet introducing these constants and this parser, none of which ever shipped, so rewrite that bullet to describe the mirror instead of adding a removal entry. The `solana-fork` entry describing the 184-byte fix in terms of byte offsets goes stale at the same time.
- The decode's minimum length requirement rises from 116 bytes to the full 184, which every live account already satisfies.

karl-dz:
Closes https://github.com/malbeclabs/doublezero/issues/4183

#### solana-sdk
- Add `ValidatorClientRewards` as a `Pod` mirror of the onchain struct, with a `PrecomputedDiscriminator` impl, a `Default` impl matching the program's, a `checked_short_description` accessor, and a compile-time assertion pinning the account at 184 bytes
- Delete the five `VCR_*_OFFSET` constants, `VCR_SHORT_DESCRIPTION_LEN`, `VCR_ACCOUNT_DATA_LEN`, the standalone discriminator constant, `parse_validator_client_rewards`, and `ValidatorClientRewardsInfo`. None of them shipped, so the changelog bullet that introduced them is rewritten rather than paired with a removal entry
- Fold the new mirror into the existing mirrored-layout section alongside `ShredRewardToken` and `ValidatorPublisherRewards` rather than opening a second one

#### solana-cli
- Read the account through the mirror in `show`, `claim`, and `init-holding`. All three decode the whole struct, so they require at least the 184 bytes the program allocates
- `claim` takes its post-transaction count from `SolanaConnection::try_fetch_zero_copy_data_with_commitment`, which warns on a missing or undecodable account where the previous code printed `(unavailable)` silently

#### solana-fork
- Build the synthetic account from `ValidatorClientRewards::default()` plus field assignment instead of copying bytes to hand-written offsets, and size the rent exemption from the built buffer. The parse-back round trip and its three `ensure!` checks go with them, being tautological against a struct the same function builds
- Rename `--synthetic-vcr-manager` to `--synthetic-validator-client-rewards-manager`, along with the fork test script and the local-validator workflow
- That rename moves the environment variable clap derives from the flag, from `SYNTHETIC_VCR_MANAGER` to `SYNTHETIC_VALIDATOR_CLIENT_REWARDS_MANAGER`. A value left under the old name is ignored rather than rejected, so the fork boots with no synthetic account and the fork test fails later at the first `validator-client-rewards show`. The flag has only ever existed under `## [Unreleased]`, so no released version is affected, but anyone tracking `main` with the old name in a local `.env` or an infra job needs to move it

#### repo
- Record the conventions this change was written under in `CLAUDE.md`: the `_key` suffix for pubkey bindings, and which helper to reach for when reading a zero-copy account

## Testing
- Unit tests across the three crates, including three cases for `checked_short_description` and an exact-match assertion pinning the rendered `show` summary byte for byte, which is what guards the format string's line continuations against a dropped escape
- The `local-validator` workflow drives `show`, `init-holding`, and `claim` against a synthetic account baked into the fork at genesis, covering the renamed flag and the mirror's field offsets end to end


crates/solana-fork/src/main.rs
```diff
@@ -75,11 +80,11 @@ struct Args {
     next_completed_dz_epoch_override: Option<u64>,
 
     /// Pubkey to use as the manager of a synthetic ValidatorClientRewards PDA
-    /// (`client_id=65535`). When set, the fork loader bakes a VCR account at
+    /// (`client_id=65535`). When set, the fork loader bakes the account at
     /// genesis so the fork test can exercise `validator-client-rewards claim`.
-    /// Defaults to disabled (no synthetic VCR).
+    /// Defaults to disabled.
     #[arg(long, env, value_name = "PUBKEY")]
```
bgm-malbeclabs, line 86: Not a blocker. Worth a line in the CHANGELOG.

`env` on this attribute derives the variable name from the field, so the rename moves it as well. It was `SYNTHETIC_VCR_MANAGER` and it is now `SYNTHETIC_VALIDATOR_CLIENT_REWARDS_MANAGER`. Anyone who still sets the old name gets no error and no synthetic account. The fork boots without the PDA, and `sh/test_doublezero_solana_fork.sh` then fails at the first `validator-client-rewards show` with nothing pointing back at the variable.

The `solana-fork` CHANGELOG entry names only the flag. Naming the environment variable there too gives anyone with the old value in a local `.env` or an infra job something to search for.

bgm-malbeclabs (commented):
Checked the change against the onchain source rather than only reading the diff.

What checks out: the new `ValidatorClientRewards` mirror in `crates/solana-sdk/src/shred_subscription/state.rs` is field for field identical to the program struct in `doublezero-shred-subscription`, including `_padding_0`, `_padding_1`, `_gap: StorageGap<2>`, and `#[repr(C, align(8))]`. Size works out to 176, so `184 - DISCRIMINATOR_LEN` holds and the offsets match the deleted `VCR_*_OFFSET` constants exactly. `discriminator_slice()` returns the same 8 bytes the old `borsh::to_vec` produced, so the synthetic fork account stays byte identical. `cargo clippy --all-targets -- -Dclippy::all` is clean on the three affected packages, unit tests pass including the exact match `show` summary assertion, and the `\` line continuations in the new `format!` reproduce the old two space indent. No references to the deleted API remain.

No correctness bug found. One note below, not a blocker.

crates/solana-fork/src/main.rs
karl-dz: Agree, and called out in the PR body too. How does this look? https://github.com/doublezerofoundation/doublezero-offchain/pull/410/commits/cc65a22ba6c747f75042439ae677023e14f07063

bgm-malbeclabs (approved):

